PR #5, opened 2015-05-14
Fix for keycloak behind proxy 

ieugen:
I have a fix for using connect-keycloak behind a proxy. To make it work you have to enable 'trust proxy' in your node app as you normally have to and keycloak will buuild the redirect url accordingly. 

This is the nginx configuration I'm using and it works with my node app.
```
location /{
                proxy_set_header Host               $host;
                proxy_set_header X-Real-IP          $remote_addr;
                proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;
                proxy_set_header X-Forwarded-Proto  $scheme;
                proxy_set_header X-Forwarded-Port   443;
                proxy_pass http://127.0.0.1:3000;
        }
```

P.S. 
It is very hard to install the modules from a github repo because you use sub-directories and npm can't handle taht [1] . If you made modifications to the modules and you wish to use them in the your deployed applications you will need npm registry or separate individual repos. 
Right now I'm using a copy of 'connect-keycloak ' in a separate github repo which is not desirable. 

[1] https://github.com/npm/npm/issues/2974

### DIFF
--- a/connect/middleware/protect.js
+++ b/connect/middleware/protect.js
@@ -1,11 +1,25 @@
 var UUID = require('./../uuid' );
 
 function forceLogin(keycloak, request, response) {
-  var host = request.hostname;
-  var port = request.app.settings.port || 3000;
+  var host = request.hostname,
+      protocol = request.protocol,
+      port, redirectPort;
+  // check to see if we are behind proxy
+  if (request.app.get('trust proxy')) {
+    port = request.headers['x-forwarded-port'];
+  } else {
+    port = request.app.settings.port || 3000;
+  }
 
-  var redirectUrl = 'http://' + host + ( port == 80 ? '' : ':' + port ) + request.url + '?auth_callback=1';
+  if( port == 80 && protocol == 'http') {
+    redirectPort = '';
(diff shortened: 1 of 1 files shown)


connect/middleware/protect.js
```diff
@@ -1,11 +1,25 @@
...
 
 function forceLogin(keycloak, request, response) {
-  var host = request.hostname;
-  var port = request.app.settings.port || 3000;
+  var host = request.hostname,
+      protocol = request.protocol,
+      port, redirectPort;
+  // check to see if we are behind proxy
+  if (request.app.get('trust proxy')) {
+    port = request.headers['x-forwarded-port'];
```
ieugen: if not behind proxy, port is undefined on this path. I  recommend 
```
 port = request.headers['x-forwarded-port'] || request.app.settings.port ;
```

ieugen:
Remaking - fixed also logout redirect url

bobmcwhirter:
Thanks. I'll either apply this or just give you commits.  Are you comfortable with that?

--

ieugen:
Yes, both ways work fine for me. Please see #6  Thanks.

PR #6, opened 2015-05-25
Connect keycloak behind reverse proxy fix

ieugen:
I have a fix for using connect-keycloak behind a reverse proxy. To make it work you have to enable 'trust proxy' in your node app (so express will process the X-Forwarder-For headers) and keycloak will build the correct redirect url. 

This is the nginx configuration I'm using and it works with my node app, deployed behind nginx+ssl .
```
location /{
                proxy_set_header Host               $host;
                proxy_set_header X-Real-IP          $remote_addr;
                proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;
                proxy_set_header X-Forwarded-Proto  $scheme;
                proxy_set_header X-Forwarded-Port   443;
                proxy_pass http://127.0.0.1:3000;
        }
```

P.S. 
It is very hard to install the modules from a github repo because you use sub-directories and npm can't handle taht [1] . If you made modifications to the modules and you wish to use them in the your deployed applications you will need npm registry or separate individual repos. 
Right now I'm using a copy of 'connect-keycloak ' in a separate github repo which is not desirable. 

[1] https://github.com/npm/npm/issues/2974

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+site/
+doc/
+node_modules/
+.grunt
+.idea
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 .grunt
 .idea
 *.swp
+npm-debug.log
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -62,6 +62,9 @@ module.exports = function(grunt) {
   });
 
   grunt.registerTask( 'finish-site', 'Finish assembling the site', function() {
+    if ( fs.existsSync(path.join('.', 'site', 'index.md' ) ) ) {
+      fs.unlinkSync(path.join('.','site','index.md'));
+    }
     fs.linkSync( path.join('.', 'README.md'), path.join( '.', 'site', 'index.md' ) );
   } );
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,77 @@
+
+var fs = require('fs');
+var path = require('path');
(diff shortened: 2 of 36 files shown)


bobmcwhirter:
I need to lobby the keycloak guys for more repos. I understand the problem. 

--

ieugen:
Thanks. I hope the repos will be splited.

Esya:
+1 Here, the connect module is unusable as is, a PR that fixed it was apparently merged yesterday but no-one republished to NPM, and I can't put it in my package.json because it is not a valid path. 

I'll need to fork it, put everything in the root folder, and then use my git repo as the source, but then I won't be able to merge upstream changes. How does this need lobbying as public repositories are free and unlimited?

bobmcwhirter:
Because I'm not the admin of the Keycloak organization. 

--

abstractj:
@stianst do you think we can split these folders into repositories to make it more Node-ish? //cc @bobmcwhirter

bobmcwhirter:
Yes, a split would be very much more Node-friendly.

Also, having just gone through this with WildFly Swarm, `git subtree` is a useful command, if you want to split sub-dirs into their own repos retaining just the portion of history for each that applies to each.

StackOverflow has the instructions, but it's not horribly difficult with a modern git (if you have `git-subtree`).

bobmcwhirter:
regarding the actual PR about reverse-proxying, it sounds like a great idea, but I have no clue about this codebase any more regarding applicability of this change.

PR #17, opened 2016-04-06
obtainDirectly, validateAccessToken supports https realm urls

rockxwre:
The `GrantManager`'s `obtainDirectly` and `validateAccessToken` functions were always using the http protocol when making Keycloak requests. Since our Keycloak server is only accepting https requests (and `auth-server-url` in keycloak.json is configured as such) this will not work. I think these functions should also support https.
To reduce some duplicate code I created a function `getProtocol` that will return the protocol instance based on an `options` argument.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+site/
+doc/
+node_modules/
+.grunt
+.idea
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 .grunt
 .idea
 *.swp
+npm-debug.log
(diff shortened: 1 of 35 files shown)


rockxwre:
Hello Bruno,

I was wondering why my pull request has been closed by you today?
Also I'm confused since the folder structure of my keycloak-nodejs-connect
fork does not match anymore with keycloak-nodejs-connect. Are the projects
keycloak-nodejs/keyclock-nodejs-connect and keycloak-nodejs-auth-utils
restructured/split up somehow today?

Thanks!
Ramon

PR #18, opened 2016-04-06
added example similar to js-console that demonstrates login/logout

winfinit:


abstractj:
@winfinit thanks, landed and tested against your changes for auth-utils

PR #19, opened 2016-04-08
Minor fixes and updates

helio-frota:
* Some jshint fixes
* Dependencies updated
* Grunt-mrdoc commented due instability of the plugin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/node_modules
+/doc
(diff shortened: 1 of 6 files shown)


lholmquist:
looks good.  

perhaps we should be adding in a .jshintrc file with the specified jshint rules?  that could be another pr too

helio-frota:
@lholmquist I agree.
@lance other day commented on the establishment of rules.

abstractj:
@lholmquist @helio-frota +1 and makes sense for another PR

@helio-frota regarding the mr-doc commented. Why not just remove it?

helio-frota:
@abstractj for me is ok, we can think later about the docs.

PR #21, opened 2016-04-08
Istanbul coverage tool added.

helio-frota:


abstractj:
@helio-frota I believe we should be consistent about between projects. auth-utils does not have any coverage tools, yet. I can be wrong, but right now I don't see any need for this. Unless you, @lance and @lholmquist are in an agreement, I'd close this PR for now and we can revisit later.

helio-frota:
@abstractj +1 can be closed.

PR #22, opened 2016-04-14
Check if the token is a string or an object before parsing

sebastienblanc:
disclaimer : I'm not a JS expert, fix could be nicer maybe. 

This extra check makes the nodejs adapter work when the application is configured as "bearer-only" , fix is inspired by this commit from the util project https://github.com/keycloak/keycloak-nodejs-auth-utils/commit/e4e8e98a051b46baf5247c7bd7d3e7c7c2275973 , not sure why it was not applied as well for this project at that time.

@abstractj @lholmquist it's all yours ... 

tested with KC 1.9.1

### DIFF
--- a/index.js
+++ b/index.js
@@ -233,12 +233,13 @@ Keycloak.prototype.getGrant = function(request, response) {
     }
   }
 
-  if (rawData) {
-    rawData = JSON.parse(rawData);
+  var grantData = rawData;
+  if (typeof(grantData)==='string') {
+    grantData = JSON.parse( grantData );
   }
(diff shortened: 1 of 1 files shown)


lholmquist:
@sebastienblanc @abstractj seems legit.  so :+1:   

i think the `var grantData` should probably be hoisted to the top of that function,  but there are other variables that needed it to,  so i wouldn't worry about it in this PR. for future clean up

abstractj:
@sebastienblanc @lholmquist thank you

PR #23, opened 2016-04-15
updates on example

helio-frota:


abstractj:
@helio-frota landed, thank you!

PR #25, opened 2016-05-02
Add function names for better debug in express

roblabla:
This makes the middleware's name show up in the debug output fo express when using `DEBUG=express:*`. Otherwise shows up as `anonymous`.


abstractj:
@roblabla could you please remove those prefixes? Ex: keycloak_adminLogout replace by just "adminLogout".

roblabla:
Done.

abstractj:
@roblabla landed, thank you

PR #26, opened 2016-05-02
Only send raw token to keep the cookie size as small as possible

roblabla:
Sending the full, parsed token makes the headers too large for most reverse proxies, which end up dropping the response. By sending only the base64'd token, the cookie size is kept at a small enough size to be sent through.

### DIFF
--- a/stores/cookie-store.js
+++ b/stores/cookie-store.js
@@ -15,7 +15,7 @@ CookieStore.get = function(request) {
 };
 
 var store = function(request, response) {
-  response.cookie( CookieStore.TOKEN_KEY, JSON.stringify( this ) );
+  response.cookie( CookieStore.TOKEN_KEY, this.__raw );
 };
 
 var unstore = function(request, response) {


abstractj:
@roblabla first, thanks for this change. Do you have the steps to reproduce the issue?

roblabla:
A minimalist example app :

```
var app = require('express')();
var Keycloak = require('keycloak');

app.use(require('cookie-parser')());
var keycloak = new Keycloak({ cookies: true });
app.use(keycloak.middleware({
  logout: '/logout',
  admin: '/'
}));
app.use(keycloak.protect());
app.use(express.static(path.join(__dirname, "..")));
app.listen(9090);
```

(Note : for this to work, you need to merge #29 too, otherwise crash on login ^^')

When running this app under the following nginx proxy :

```
server {
  listen 80;
  listen [::]:80;
  location / {
    proxy_pass http://localhost:9090;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
  }
}
```

NginX will return a 502 BAD Gateway with the following message in the logs :

```
May 13 13:45:43 nixos nginx[23173]: nixos nginx: 2016/05/13 13:45:43 [error] 23173#23173: *2094 upstream sent too big header while reading response header from upstream, client: 92.163.56.222, server: x.cmc.im, request: "GET / HTTP/1.1", upstream: "http://127.0.0.1:9090/", host: "x.cmc.im"
```

abstractj:
@roblabla gotcha, thank you!

PR #27, opened 2016-05-03
test with jshint.

helio-frota:


abstractj:
@helio-frota like we discussed, let's run/fix jshint issues before we add such constraints to the project.

abstractj:
@helio-frota I'm closing it for now. I believe we have to coordinate with @lance to should include checks for caMelCasE or not. Maybe, at least keep it in sync with https://github.com/keycloak/keycloak-nodejs-auth-utils/blob/master/.jshintrc

PR #28, opened 2016-05-03
Use token.hasRole to check for roles of other apps

roblabla:
This allows using `<appname>:<role_name>`. Note that it might break for those that use `app:<role_name>`.

### DIFF
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -16,14 +16,7 @@ function forceLogin(keycloak, request, response) {
 }
 
 function simpleGuard(role,token) {
-  if ( role.indexOf( "app:" ) === 0 ) {
-    return token.hasApplicationRole( role.substring( 4 ) );
-  }
-  if ( role.indexOf( "realm:" ) === 0 ) {
-    return token.hasRealmRole( role.substring( 6 ) );
-  }
-
-  return false;
+  return token.hasRole(role);
 }
 
 module.exports = function(keycloak, spec) {


abstractj:
@lance I think we're good to merge this, once we already check it, right here https://github.com/keycloak/keycloak-nodejs-auth-utils/blob/master/lib/token.js#L79-L93.

What do you think?

lance:
LGTM

abstractj:
@roblabla landed here https://github.com/keycloak/keycloak-nodejs-connect/commit/83bea17aff9f4c3d6adcdf88b3f1661be9b92c15. Thank you!

jimdillon:
Good fix roblabla, I ran into the same issue, seems like npm hasn't picked up the master branch that contains the fix, so I'm going to pull from the master branch.

@lance do you know when this patch might be reflected in a tag?  I'd much rather pull from a known tag than a branch that will change.

lance:
@jimdillon that question is probably best left to @abstractj. I think both this module and `keycloak/keycloak-nodejs-auth-utils` could use a release some time in the near future, and that process is controlled by @abstractj, not me.

jimdillon:
@lance, thank you for the information, @abstractj what do you think, what does the release timeline look like?  I'm working on a Red Hat Mobile demo using this great piece of middleware and would like to be able to let npm do it's thing rather than set a specific branch or commit for the package.

PR #29, opened 2016-05-13
Remove store.getId

roblabla:
`store.getId` isn't implemented in cookie store, leading to a crash when logging in. I'm not really sure what its use is anyway, so I figured I could probably just delete it...


roblabla:
Bump. This is still a pretty big problem. Cookie store is pretty much unusable without this fix :(. Any news ?

lance:
LGTM

PR #34, opened 2016-06-07
Changed redirect URL to use originalUrl instead of url

gaganinvic:
Changed redirect URL to use originalUrl instead of url as it was stripping the baseUrl

### DIFF
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -6,7 +6,7 @@ function forceLogin(keycloak, request, response) {
   var port = headerHost[1] || '';
   var protocol = request.protocol;
 
-  var redirectUrl = protocol + '://' + host + ( port === '' ? '' : ':' + port ) + request.url + '?auth_callback=1';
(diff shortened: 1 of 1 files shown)


middleware/protect.js
```diff
@@ -6,7 +6,7 @@ function forceLogin(keycloak, request, response) {
   var port = headerHost[1] || '';
   var protocol = request.protocol;
 
-  var redirectUrl = protocol + '://' + host + ( port === '' ? '' : ':' + port ) + request.url + '?auth_callback=1';
+  var redirectUrl = protocol + '://' + host + ( port === '' ? '' : ':' + port ) + request.originalUrl + '?auth_callback=1';
```
lance: I think this should be `(request.originalUrl || request.url)` to maintain backwards compatibility with pre 4.x express.

lance:
@gaganinvic thanks!

PR #43, opened 2016-06-29
remove this and make the store function functional

lholmquist:
@lance @abstractj  @helio-frota mind taking a quick look.

For https://issues.jboss.org/browse/KEYCLOAK-3199

instead of adding the function keyword back in, i've made it more "functional" so its not as confusing where this comes from

### DIFF
--- a/stores/cookie-store.js
+++ b/stores/cookie-store.js
@@ -30,8 +30,10 @@ CookieStore.get = (request) => {
   }
 };
 
-let store = (request, response) => {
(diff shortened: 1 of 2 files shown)


lance:
LGTM - any chance you could add a unit test of some kind?

lholmquist:
@lance oh, right,  i was going to do that

lholmquist:
@lance this is turning out to be a little tricky to test,  i'm going to probably merge this, and do a release so things are breaking anymore,  then go back and do the tests

lance:
+1

PR #45, opened 2016-07-08
Update example to version 0.2.1

foo4u:
Example version still infinitely redirects. Update example to 0.2.1.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -9,7 +9,7 @@
   "author": "Roman Jurkov <winfinit@gmail.com>",
   "license": "Apache-2.0",
   "dependencies": {
-    "keycloak-connect": "0.2.0",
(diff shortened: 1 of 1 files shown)


foo4u:
Alternatively what if we set the example to refer to the parent directory for Keycloak Connect:

```
"keycloak-connect": "file:../",
```

abstractj:
@foo4u +1 to the second alternative. @lance @lholmquist what do you think?

helio-frota:
@foo4u good, this solves the problem with keep updating the example
+1
also better to update the doc about `npm link`.

foo4u:
Ok, will update the PR.

foo4u:
Going to close this as I see a commit in master that uses the parent folder as the source for Keycloak Connect.

PR #47, opened 2016-07-28
registering the unprotected endpoint before registering the KC middle…

lholmquist:
…ware

i've moved the registration of the unprotected endpoint '/' to be before any of the KC middleware gets added.  This way none of the KC functions will be called when accessing the '/' endpoint, since it is unprotected and there would be no need for any KC functions to be called.

@abstractj mind taking a look.  this is more of a design issue than anything, so not to important :)


sebastienblanc:
:+1:

abstractj:
@lholmquist landed, thank you

PR #49, opened 2016-08-01
KEYCLOAK-3384: Fix login redirect URL generation

foo4u:
Check if original request already contains a query string and append auth_callback=1 to existing query string if one exists; add query string otherwise. Fixes [KEYCLOAK-3384][0].

[0]: https://issues.jboss.org/browse/KEYCLOAK-3384


sebastienblanc:
Maybe you could add a test case in the test script.

foo4u:
@sebastienblanc will add tests shortly and update.

foo4u:
Rebased on master and added tests.

abstractj:
@foo4u thank you, landed

PR #50, opened 2016-08-01
KEYCLOAK-3387: Add redirect to login function

foo4u:
Add an overridable function, Keycloak.prototype.redirectToLogin(), to determine if an unauthorized request should be redirected to Keycloak for login or if an access denied error should be returned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,22 @@ for each section:
     }
 
     app.get( '/:section/:page', keycloak.protect( protectBySection ), sectionHandler );
+    
+### Advanced Login Configuration
+
+By default, all unauthorized requests will be redirected to the Keycloak login
+page unless your client is bearer-only. However, a confidential or public client
+may host both browsable and API endpoints. To prevent redirects on unauthenticated
+API requests and instead return an HTTP 401, you can override the `redirectToLogin` 
+function.
+
+For example, this override checks if the url contains /api/ and disables login
+redirects:
+
+	Keycloak.prototype.redirectToLogin = function(req) {
+      var apiReqMatcher = /\/api\//i;
+      return !apiReqMatcher.test(req.originalUrl || req.url);
+    };
 
 ## Additional URLs
 
--- a/index.js
+++ b/index.js
@@ -322,4 +322,8 @@ Keycloak.prototype.getAccount = function (token) {
   return this.grantManager.getAccount(token);
 };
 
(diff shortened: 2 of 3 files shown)


foo4u:
This PR truthfully only addresses part of KEYCLOAK-3387. It allows the app developer to determine when a redirect is needed by overriding `redirectToLogin()` but it should be enough for most users.

For example, this override checks if the url contains /api/ and disables redirects to login:

```
Keycloak.prototype.redirectToLogin = function(req) {
    var apiReqMatcher = /\/api\//i;
    return !apiReqMatcher.test(req.originalUrl || req.url);
};
```

We could try to do some magic like checking if `X-Requested-With` header is set to `XMLHttpRequest` to make it smarter by default but I still think this decision needs to be exposed to the developer in some fashion.

sebastienblanc:
Interesting, so this is to deal with a hybrid situation ? I wonder how the other adapters manage (or not) this.

foo4u:
@sebastienblanc well, I wrote the Keycloak Spring adapter and we also had this requirement from a number of users. By default the [`KeycloakAuthenticationEntryPoint`][0] uses a [header check][1] to determine if it should redirect or return an HTTP 401.

However, if this doesn't suit a customer's use case, they can extend or replace our security entry point to provide their own logic. The main thing of importance here is to make it customizable.

[0]: https://github.com/keycloak/keycloak/blob/5c98b8c6ae7052b2d906156d8fc212ccd9dfd57d/adapters/oidc/spring-security/src/main/java/org/keycloak/adapters/springsecurity/authentication/KeycloakAuthenticationEntryPoint.java
[1]: https://github.com/keycloak/keycloak/blob/5c98b8c6ae7052b2d906156d8fc212ccd9dfd57d/adapters/oidc/spring-security/src/main/java/org/keycloak/adapters/springsecurity/authentication/HttpHeaderInspectingApiRequestMatcher.java

sebastienblanc:
@foo4u Yes this make sense and I think the override is the good balance. Could you maybe add your override example to the readme ? 
Otherwise I'm :+1:

foo4u:
Will update README.

foo4u:
@sebastienblanc I updated the README as requested and rebased on master.

sebastienblanc:
:+1: @abstractj  ok to merge ?

abstractj:
@foo4u thank you, landed

PR #52, opened 2016-09-02
KEYCLOAK-3516 Better test organization.

helio-frota:
We need to move unit tests related to
keycloak javascript object toanother
directory:

tests/unit/keycloak-object-test.js

This will make things more clear.


abstractj:
@helio-frota landed, thank you

PR #63, opened 2016-11-29
[KEYCLOAK-3284]: Web browser SSO with Node.js adapter does not work

abstractj:


example/nodejs-example-realm.json
```diff
@@ -47,6 +47,8 @@
             "enabled": true,
             "publicClient": true,
             "baseUrl": "/",
```
vmuzikar: Not sure if there should be two `baseUrl` elements for the same client. It's working like that but I'd remove this one.

PR #65, opened 2016-12-01
add support for not before endpoint and refactoring the admin script

sebastienblanc:
This PR adds the feature to "listen" to "not_before" requests sent by te admin console. 
There is also refactoring of the admin-logout script (renamed to admin) to be more generic and to be make it easy in the future to add other admin endpoints available. 

@abstractj Do you mind to review ?


abstractj:
@sebastienblanc landed here https://github.com/keycloak/keycloak-nodejs-connect/commit/4f99a8c582eb53c71f220a54e3667b259d9c5e24, thank you.

PR #67, opened 2016-12-07
Update auth-utils dependency

abstractj:
@sebastienblanc could you please review?


sebastienblanc:
:+1: As long we coordinate well the npm releases for both libs ;)

abstractj:
@sebastienblanc both modules are released together, so I don't think we gonna hit any issues.

PR #69, opened 2017-02-07
Do not force strict mode in javascript files

strk:
You can use --use_strict switch from Node (0.10+) and latest releases
(9.5+ tested) have it on by default.

See issue KEYCLOAK-4387

### DIFF
--- a/middleware/admin.js
+++ b/middleware/admin.js
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-'use strict';
+
 
 function Admin (keycloak, url) {
   this._keycloak = keycloak;
--- a/middleware/grant-attacher.js
+++ b/middleware/grant-attacher.js
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations under
  * the License.
  */
-'use strict';
+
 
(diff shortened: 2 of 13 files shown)


stianst:
Quick Google search suggests that using strict is recommended, so you'll have to explain in more detail why it should be removed.

helio-frota:
+1 to keep `strict mode`.

> 1.strict mode eliminates some JavaScript silent errors by changing them to throw errors.
> 2. strict mode fixes mistakes that make it difficult for JavaScript engines to perform optimizations: strict mode code can sometimes be made to run faster than identical code that's not strict mode.
> 3. strict mode prohibits some syntax likely to be defined in future versions of ECMAScript.

(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode)

strk:
See the ticket (KEYCLOAK-4387) for the rationale of the suggested change.
I'd be happy if it worked while in strict mode.

stianst:
Seems like there's a issue in keycloak-object-test that needs fixing rather than changing to strict mode.

strk:
Ok, I'll close this then. Will file more tickets for other issues (ie: not finding, in documentation, example of the "require" call to use)

PR #72, opened 2017-02-17
Encode minimum required node version

strk:
See http://lists.jboss.org/pipermail/keycloak-user/2017-February/009574.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,9 @@
     "semistandard": "^9.2.1",
     "tape": "^4.6.3"
   },
+  "engines" : {
+    "node" : ">=4.0.0"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/keycloak/keycloak-nodejs-connect.git"


iperdomo:
This change is failing because a problem in the build pipeline. It tries to get the published version from the Keycloak website.

    curl -s http://www.keycloak.org | grep -i version | head -n1 | grep -o "'.*'" | sed -e "s/'//g"

The above command is returning an empty string.

I think is safer to get the latest version from the published distribution in Maven?

````
$ curl -s -H "Accept: application/json, text/javascript, */*; q=0.01" -H "Accept-Language: en-US,en;q=0.5" 'https://search.maven.org/solrsearch/select?q=g:org.keycloak+AND+a:keycloak-server-dist&rows=20&core=gav&wt=json' | jq .response.docs[0].v

"2.5.1.Final"
````

This has the requirement of `jq` for processing the JSON response

iperdomo:
Well it seems that @abstractj came with a similar solution (checking Maven) without requiring an extra dependency:

https://github.com/keycloak/keycloak-nodejs-connect/pull/71/commits/bd37f52e868b3d289f3ae966d37f389d6ddd228b#diff-9257b6f4640537b3c313e06dcdb250a5R3

abstractj:
Landed here https://github.com/keycloak/keycloak-nodejs-connect/commit/40b84de5e92b45e9f1a048c5145d5ea03cc580a1. Thank you.

PR #74, opened 2017-02-22
specifiy keycloak-auth-utils dependency

mdunisch:
Instead of linking to github I changed dep to npm-package.
- In our corporate firewall github is blocked
- Requesting a Github without specific version is dangeros

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,11 +33,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "keycloak-auth-utils": "keycloak/keycloak-nodejs-auth-utils"
+    "keycloak-auth-utils": "^2.5.1"
   },
-  "bundledDependencies": [
-    "keycloak-auth-utils"
-  ],
   "devDependencies": {
     "express": "^4.14.0",
     "express-session": "^1.14.2",


abstractj:
I suggest you to grab the latest released version from https://www.npmjs.com/~keycloak. The master branch is used exclusively for development, which means that both `keycloak-nodejs-connect` and `keycloak-auth-utils` must reflect the latest changes on GitHub.

What do you mean "Requesting a Github without specific version is dangerous"? Could you please elaborate?

mdunisch:
Hey,

yes we used the latest release from https://www.npmjs.com/package/keycloak-connect but in it's dependencies this package requests the github master branch from `keycloak-auth-utils`.

**Why that's dangerous**
Npm let me define a specific version of a package I want to use. For a production environment I can install this version and always get the same result. When you refer to the latest master branch instead of a npm package the result of my npm install can change every time. For example if I target `keycloak-connect: 0.1.16` it's possible that in year I get a complete different and incompatible `keycloak-auth-utils` when installing and this would break my production environment.

Thats why you should use npm package dependencies - maybe with different version range.

abstractj:
@mdunisch I see your point. But I'd rather to change it dynamically during prepublish, maybe? @lance @lholmquist do you have any thoughts/best practices about it?

abstractj:
@mdunisch thank you for your PR. I'm closing it, because I believe this change https://github.com/keycloak/keycloak-nodejs-connect/pull/75/files fixes the issue. Feel free to comment on that if you disagree.

mdunisch:
Hi @abstractj ,
thank you that fixes the "get latest from github"-Problem. But I still think that doesn't solve the problem with incompatible versions. It's still the same: If I install a specific version of keycloak-nodejs-connect it's not sure that  "keycloak-auth-utils": "*" will solve a compatible dependency. 

You should define a version (or at least main-versions) in your package.json so make a consistency npm-package.

abstractj:
@mdunisch I was thinking about npm update. But you're right, you cannot guarantee. I will add some changes to our release script, to make sure that both modules are following the same and stable version.

Thanks for your input.

PR #76, opened 2017-03-29
Allow usage behind load balancer

vanetix:
This allows a nodejs application running behind a load balancer (with ssl termination) to still negotiate the proper protocol to use with the keycloak backend.

### DIFF
--- a/middleware/logout.js
+++ b/middleware/logout.js
@@ -30,7 +30,8 @@ module.exports = function (keycloak, logoutUrl) {
     let host = request.hostname;
     let headerHost = request.headers.host.split(':');
     let port = headerHost[1] || '';
-    let redirectUrl = request.protocol + '://' + host + (port === '' ? '' : ':' + port) + '/';
+    let protocol = request.headers['x-forwarded-proto'] || request.protocol;
+    let redirectUrl = protocol + '://' + host + (port === '' ? '' : ':' + port) + '/';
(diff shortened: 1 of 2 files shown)


stianst:
Can you create a JIRA for this please?

What about X-Forwarded-Host? Also, handling X-Forwarded headers should be an option that has to be enabled as otherwise the headers can be used maliciously when there isn't a load balancer in front.

vanetix:
Sure, here is the original issue in JIRA: https://issues.jboss.org/browse/KEYCLOAK-4334

I'd be happy to update if this seems like a feasible request? This might even be better suited for the `"ssl-required"` (isn't used presently) option in the keycloak.json configuration.

stianst:
Would be good to have the same behaviour as the Java adapters:
* X-Forwarded* headers are used, but needs an option to be enabled
* SSL required option will be happy if https is used to the proxy and won't require proxy to Keycloak to use https

abstractj:
@vanetix I believe it's a good start. Would you like to add the changes mentioned by Stian?

vanetix:
Sure - Let me take a look at the java adapter and I'll make the configuration options the same to try and keep everything inline.

abstractj:
@vanetix I'm closing it for now, until the PR gets updated. But feel free to reopen as soon as you get more changes to be reviewed, please.

PR #102, opened 2017-11-30
Add external-url configuration settting

darabos:
For https://issues.jboss.org/browse/KEYCLOAK-5954.

This setting makes it possible to override the scheme/hostname/port of the redirection. Very useful if your server is behind a reverse proxy.

### DIFF
--- a/middleware/auth-utils/config.js
+++ b/middleware/auth-utils/config.js
@@ -161,6 +161,12 @@ Config.prototype.configure = function configure (config) {
     }
     this.publicKey += '-----END PUBLIC KEY-----\n';
   }
+
+  /**
+   * External URL of the server. Only needed if behind a reverse-proxy.
+   * @type {String}
+   */
+  this.externalUrl = resolveValue(config['external-url'] || config.externalUrl);
 };
 
 module.exports = Config;
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -24,7 +24,8 @@ function forceLogin (keycloak, request, response) {
   let protocol = request.protocol;
   let hasQuery = ~(request.originalUrl || request.url).indexOf('?');
 
-  let redirectUrl = protocol + '://' + host + (port === '' ? '' : ':' + port) + (request.originalUrl || request.url) + (hasQuery ? '&' : '?') + 'auth_callback=1';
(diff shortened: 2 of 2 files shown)


abstractj:
@darabos Thanks for the PR. We need to have tests added for this.

We can also not accept any more changes/enhancements for Keycloak 3.x as we are currently in feature freeze and won't be able to merge this until we start development on Keycloak 4.x

darabos:
> @darabos Thanks for the PR. We need to have tests added for this.

Sure. I just wanted to check with you if this was even a good idea before spending time on the tests.

> We can also not accept any more changes/enhancements for Keycloak 3.x as we are currently in feature freeze and won't be able to merge this until we start development on Keycloak 4.x

All right, thanks for the info! Is there an issue I can watch to know when the freeze is over? I'd be fine with letting this linger until 4.x starts.

abstractj:
@darabos sure thanks for your interest. You can look for bug fixes, but being completely honest these fixes will only get in on the next release, into other words, next year.

Flowkap:
I searched some time why it wasn'T working behind reverse proxy with path routing. Thought I did misconfigure express. Anyway found out, that keycloak builds the callbacks fully qualified hence relative linking does not work at all. 

I'll try out the PR if it's sufficient for me. Another problem I found is that external https was not preserved although set in express. And a relative authUrl produced invalid tokens. I'll check both with the PR code.

abstractj:
@darabos could you please add tests?

darabos:
Yeah, I can try to add tests later this week. But first I'll try if I can just use the right headers to do the same like @Flowkap says. If that works (it should) then I can change the pull request to update the docs instead of the code. 😄  Thanks!

abstractj:
@darabos closing it. But feel free to reopen if you feel like this change should be included.

korommatyi:
The solution @Flowkap proposed almost works. By using `x-forwarded-proto` and `x-forwarded-host` the protocol and the host is correctly identified, but the port is still not the port of the proxy server since it's determined based on [`request.headers.host`](https://github.com/keycloak/keycloak-nodejs-connect/blob/a71dea163a4818816fe2f4f1276ed29940bdad47/middleware/protect.js#L22-L23) and not `x-forwarded-port`! :sob:

PR #103, opened 2017-12-06
KEYCLOAK-5823 Fix access token automatic refresh

foo4u:
Attempt to refresh the access token before doing grant validation if the client is not bearer-only.

### DIFF
--- a/middleware/auth-utils/grant-manager.js
+++ b/middleware/auth-utils/grant-manager.js
@@ -241,28 +241,39 @@ GrantManager.prototype.createGrant = function createGrant (rawData) {
   let grantData = rawData;
   if (typeof rawData !== 'object') grantData = JSON.parse(grantData);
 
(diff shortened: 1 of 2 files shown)


abstractj:
@foo4u thanks for your PR, but I'm afraid that this must wait.

We cannot accept any more changes/enhancements for Keycloak 3.x as we are currently in feature freeze and won't be able to merge this until we start development on Keycloak 4.x

foo4u:
Ok, it’s a fairly major bug in the 3.x line but I’m in no rush. Just forked the repo and applied to our code base.

abstractj:
@foo4u we got an extra week for it and I managed to look at this issue today. Landed here https://github.com/keycloak/keycloak-nodejs-connect/commit/68dfa1154a061905f579d24f7169acca54eef153. Thank you.

PR #106, opened 2018-01-02
KEYCLOAK-5823 Fix service calls

foo4u:
Fix regression causing service calls to clients not configured as bearer-only to fail.

This change makes Keycloak Connect behave similarly to Keycloak's Java adapters when a token is requested and but a refresh token is not present.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 node_modules/
 keycloak-*
 !keycloak-*.json
+!keycloak-connect-rest-mixed-client-spec.js
 !keycloak-connect-test.js
 !keycloak-object-test.js
 !keycloak-fixture.json
--- a/index.js
+++ b/index.js
@@ -254,7 +254,6 @@ Keycloak.prototype.getGrant = function (request, response) {
   if (grantData && !grantData.error) {
     var self = this;
     return this.grantManager.createGrant(JSON.stringify(grantData))
-    .then(grant => { return this.grantManager.ensureFreshness(grant); })
     .then(grant => {
       self.storeGrant(grant, request, response);
       return grant;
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 node_modules/
 keycloak-*
 !keycloak-*.json
+!keycloak-connect-rest-mixed-client-spec.js
```
foo4u: I only unignored this test class for b/c unignoring all causes format errors to show up that were hidden and don't want to bloat this PR. I will send another PR cleaning up the other tests.

index.js
```diff
@@ -254,7 +254,6 @@ Keycloak.prototype.getGrant = function (request, response) {
   if (grantData && !grantData.error) {
     var self = this;
     return this.grantManager.createGrant(JSON.stringify(grantData))
-    .then(grant => { return this.grantManager.ensureFreshness(grant); })
```
foo4u: This is already done properly in `createGrant()`.

.gitignore
abstractj: @foo4u I appreciate your changes, but I don't think we should ignore it. Leave it behind will just create a technical debt for people to fix our CI in the future. We don't need to rush into things.

abstractj:
@foo4u could you please rebase your changes against master and squash your commits into a single one?

.gitignore
abstractj: @foo4u ignore this comment please. I overlooked this change.

foo4u: No problem, I’ll fix the formatting errors in the other tests today then and push another commit.

foo4u:
Yes will do.

.gitignore
foo4u: Lol, commented too quickly myself. Will fix tests in another PR as planned.

abstractj: @foo4u which tests exactly are broken? I couldn't find any broken tests or formatting issues in your PR.

foo4u: No issues in my PR @abstractj ... problem is that some test files are in the Git but also in `.gitignore` and are therefore not checked by semistandard but do contain formatting issues.

foo4u:
@abstractj rebased and squashed. Thanks!

abstractj:
@foo4u landed, thank you

PR #115, opened 2018-02-15
Update post-auth.js

PavelPolyakov:
While using this middle layer with https://www.fastify.io, I found that if no query is sent, the `request.query` is undefined.
Please, update the package if possible.

### DIFF
--- a/middleware/post-auth.js
+++ b/middleware/post-auth.js
@@ -19,7 +19,7 @@ const URL = require('url');
 
 module.exports = function (keycloak) {
   return function postAuth (request, response, next) {
-    if (!request.query.auth_callback) {
+    if (!request.query || !request.query.auth_callback) {
       return next();
     }
 


PavelPolyakov:
okay, the reason is not really the one which is stated. the thing is that `keycloak-nodejs-connect` uses `req`, `res` from `express`. While `fastify` provides the original nodejs ones.

I believe that original `connect` also uses the one from node. It will be nice if `express` specifics are filtered out, but, probably, this wouldn't happen anytime soon.

abstractj:
@PavelPolyakov please file a Jira with a description and the steps to reproduce what this PR fixes and attach your PR. Thanks in advance.

PavelPolyakov:
Hello @abstractj ,

actually this one doesn't help, only makes it safer. But afterwards I've discovered that the whole module is not really `connect` one, as I see, more `express` one.

And I created an issue for this:
https://issues.jboss.org/browse/KEYCLOAK-6625

Therefore I'll close this one.

PR #116, opened 2018-02-24
Display promise error massage

onigra:
I was hard to debug in `GrantManager.prototype.validateToken`. 
(The reason was self signed certificate error)

So, I want to display promise error reason. 🙏


onigra:
Woops, I should have created JIRA Issue.

PR #118, opened 2018-03-03
[KEYCLOAK-4162] Multi-tenancy support

jderuere:
- Enable to pass an array of config, one per realm
- When using multiple configs, a resolver that fetch the realm name from the request needs to be provide
- Retrocompatible, it still works as usual


jardakotesovec:
Hi, it would be awesome to have this feature baked in so no need for [alternative packages](https://github.com/devsu/keycloak-nodejs-multirealm), is something blocking it?

jardakotesovec:
Please let me know if there is something I can help with to make this PR happen :-).

paulmowat:
Would be great to have this feature. Ideal for a new project I'm just about to start.

jardakotesovec:
@paulmowat Hi, I shared my approach to handle multiple realms with existing keycloak-connect with simple wrapper in [jira](https://issues.jboss.org/browse/KEYCLOAK-4162). So far I think its good solution, but we are also in initial stage of the project, so time will tell if there are some issues with this approach.

comply-online:
any news about this? It would be a great addition

abstractj:
@jderuere thanks for putting this together and sorry if I didn't get back to you earlier. We are approaching code freeze and have been prioritizing bugs, instead of new features. For now I'm putting this change on "hold". Even thoughthis change is a great addition to the Node.js adapter, it also requires more testing, just to make sure that everything is in place.

abstractj:
@jderuere thanks for submitting this. It's been a while since we talked and this PR definitely need some upgrade. Even though I believe this would be something interesting for the adapter, this is a feature that introduces considerable changes, for that, we really need integration tests instead of unit tests, plus, update the documentation.

For now I'm closing this, but if you would like to add the integration tests and docs, please reopen this pull-request. Otherwise, in the future we gonna take into consideration your changes and give the proper credits.

@jardakotesovec @paulmowat @comply-online I appreciate your interest in this change, but in order to keep quality, we need to provide integration tests and documentation. If some of you are interested on it, feel free to work together with @jderuere.

jardakotesovec:
@abstractj No problem, the approach I illustrated in [jira](https://issues.jboss.org/browse/KEYCLOAK-4162) works just fine for us.. so I don't think there is real need to support that inside the middleware itself.

PR #120, opened 2018-04-05
Keycloak 6073

rightisleft:
Given that keycloak services may operate on different DNS (IE: docker services on overlay networks, and web browsers on hosts networks) - then a secured backend client should be able to explicitly declare a white listed external domain realm.

https://issues.jboss.org/browse/KEYCLOAK-6073

### DIFF
--- a/middleware/auth-utils/config.js
+++ b/middleware/auth-utils/config.js
@@ -135,6 +135,13 @@ Config.prototype.configure = function configure (config) {
    * @type {String} */
   this.realmAdminUrl = this.authServerUrl + '/admin/realms/' + this.realm;
 
+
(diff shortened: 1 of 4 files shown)


stianst:
This only addresses one small part of the problem layed out in KEYCLOAK-6073. We need a proper solution to the issue not this small workaround.

I'm updating the JIRA to clearly explain what needs to happen for this to be resolved.

stianst:
I'm going to reject this as it's not a correct solution for KEYCLOAK-6073.

rightisleft:
@stianst - is there a way to make this forward compatible with the work being done on core?

PR #121, opened 2018-04-05
KEYCLOAK-6909 / KEYCLOAK-6986 Upstream product build stuff

ASzc:
This adds the build POM and assembly XMLs from the downstream repo, with some
improvements to the POM. Notably, [jsonpath-maven-plugin](https://github.com/ASzc/jsonpath-maven-plugin) and PME replace the
overly hacky ant string replacement logic for package.json

As this is only intended for building the product, there aren't community and
product profiles.

Since the product build pom depends on mvn properties, I have included
prod-arguments.json from KEYCLOAK-6909 in this same commit. This will allow a
build pipeline to substitute the appropriate values for things like the NPM
registry.

The community build is unaffected by this change, except for a modification to
nsp-check in package.json. I don't actually know what the impact of this change
is, or why it wasn't upstreamed before.


stianst:
Looks good to me, but please let @abstractj and @martin-kanis take a look before merging

package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "jshint *.js stores/*.js middleware/*.js middleware/auth-utils/*.js example/*.js",
     "format": "semistandard",
     "test": "tape test/unit/*.js test/*.js",
-    "prepublish": "nsp check",
+    "nsp-check": "nsp check",
```
martin-kanis: This was a "hack" added to product repository when we migrated to PNC. "nsp check" downloads a lot of stuff from internet which is blocked by firewall in product build resulting in build failure. Changing "prepublish" to "nsp-check" (non-existing phase) basically turned off this check. 
On the other hand I think we want this check in community builds but this change turns it off.

ASzc:
Oh, and regarding multiple repos or branches, I don't think that would work for PNC.

The assumption is your source for a particular build are contained in a single checkout/clone. Source includes all the build files in addition to the actual source code.

ASzc:
To keep nsp-check in community, but disabled in product, I've removed the nsp change from this commit, and added another PME instruction, to delete the prepublish phase:

```diff
diff --git a/package.json b/package.json
index 1cd0867..2dbbc60 100644
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lint": "jshint *.js stores/*.js middleware/*.js middleware/auth-utils/*.js example/*.js",
     "format": "semistandard",
     "test": "tape test/unit/*.js test/*.js",
-    "nsp-check": "nsp check",
+    "prepublish": "nsp check",
     "docs": "jsdoc --verbose -d docs -t ./node_modules/ink-docstrap/template -R README.md index.js ./middleware/*.js stores/*.js ./middleware/auth-utils/*.js",
     "coverage": "istanbul cover tape test/unit/*.js tape test/*.js"
   },
diff --git a/prod-arguments.json b/prod-arguments.json
index 3462421..3748f93 100644
--- a/prod-arguments.json
+++ b/prod-arguments.json
@@ -14,7 +14,7 @@
         "dependencyManagement": "org.jboss.eap:jboss-eap-parent:$EAP_VERSION",
         "dependencyRelocations.org.wildfly:@org.jboss.eap:": "$EAP_VERSION",
         "dependencyExclusion.org.jboss:jboss-parent@*": "19.0.0.redhat-2",
-        "jsonUpdate": "package.json:$.devDependencies.keycloak-admin-client:^0.12.0"
+        "jsonUpdate": "package.json:$.devDependencies.keycloak-admin-client:^0.12.0,package.json:$.scripts.prepublish:"
     }
   }
 }
```

abstractj:
I totally understand the motivation to use Maven. I just wonder if we can provide separated branches for it.

> Oh, and regarding multiple repos or branches, I don't think that would work for PNC.
> 
> The assumption is your source for a particular build are contained in a single checkout/clone. Source includes all the build files in addition to the actual source code.

Can't you do something like git clone -b nodejs-product `git@github.com:keycloak/keycloak-nodejs-connect.git`?

If there's not other way, it's ok.

ASzc:
PNC is supposed to have network restrictions to block accessing arbitrary internet resources. I'm not sure if its actually implemented, but even if it's not, doing so would break build reproducibility.

PNC doesn't build directly from GitHub, the source for each build is archived internally.

stianst:
I would like to have everything in the same repo. I would suggest moving the prod specific files out of the root directory and into a folder, for instance 'product'. That will help to not clutter the root directory and also make it more obvious what the files are.

abstractj:
> I would like to have everything in the same repo. I would suggest moving the prod specific files out of the root directory and into a folder, for instance 'product'. That will help to not clutter the root directory and also make it more obvious what the files are.

That would be fantastic +1

ASzc:
This is doable I think. It makes the pipeline scripts much simpler to write with `prod-arguments.json` consistently in the repo's root for all builds, but it shouldn't be too hard to change that.

I will push a replacement commit soon.

ASzc:
Pushed a replacement commit with all the files in `product/`

I've also added code to the pipeline so it can handle this within itself, and when configuring PNC's alignment and building steps.

PR #123, opened 2018-04-20
Add support for TypeScript via d.ts file

evanshortiss:
I've opened this to generate some discussion regarding whether this could be supported since I use TypeScript frequently and find it great.

Supporting types is done by either adding a `d.ts` file to the repository itself or publishing them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/) so users can optionally install them via `npm install @types/keycloak-connect`.

Shipping them with the module is convenient since they come bundled, but it arguably adds overhead to releases since they need to be kept in sync with the JavaScript API - if you change a publicly accessible function, property, argument, return val etc. then the typings need an update meaning possibly more  patch releases if things are missed. If they're shipped separately you can remove this worry to an extent since it's not officially part of the module.


keycloak.d.ts
```diff
@@ -0,0 +1,348 @@
...
+
+  interface GrantProperties {
+    access_token?: string
+    refresh_token?: string
+    id_token?: string
+    expires_in?: string
+    token_type?: string
+  }
+
+  interface Token {
```
evanshortiss: Technically these _should_ be classes, but it causes lines 11 and 12 not work, so using interfaces is a decent compromise.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "4.0.0-beta.1",
   "description": "Keycloak Connect Middleware",
   "homepage": "http://keycloak.org",
-  "main": "index.js",
+  "main": "keycloak.js",
```
evanshortiss: Not 100% sure this is required to match for the `d.ts` to work, but no harm in doing so.

ssilvert:
I don't really know about the Node.js adapter, but I think the use case is the same as we have for the keycloak.js adapter.  If you are using TypeScript then you want type safety and IDE code completion features.

Every javascript library these days should bundle a *.d.ts file.  It doesn't add TypeScript or affect the runtime in any way.  It just allows TypeScript developers to have a better experience with your library.

abstractj:
@evanshortiss also, please file a Jira

evanshortiss:
@abstractj [here's](https://issues.jboss.org/browse/KEYCLOAK-6409) an existing JIRA I commented on related to this.

As for the use case it enables support for TypeScript to offer a ( subjectively 😉 ) better development experience via compile time type checks. Even developers using regular JavaScript can benefit if their editor supports reading the files as @ssilvert pointed out since they will get the type ahead. So it's not strictly necessary; it simply provides better development experience for JavaScript and TypeScript developers.

Below is a GIF showing what it looks like in action. You can see that input params, functions, properties, etc. all appear in intellisense for my TypeScript project.

ts-keycloak

It's possible to not include this directly in this repository and instead have it supported upstream in [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/). Doing this means the typings will more easily become out of sync with this repo since they would not be officially maintained, but it's not a huge issue to PR them. Also, if they types are in DefinitelyTyped then developers would need to install separately via `npm i @types/keycloak-connect -D`

Hope this helps 😄

abstractj:
@evanshortiss thanks Evan. Just updated that Jira with your PR. @ssilvert would you mind to look at this when you get the chance?

keycloak.d.ts
ssilvert: For my own edification, why do you say they should be classes?

MikeyBurkman:
One thing that should be noted is that VS Code will utilize the typings for intellisense and autocomplete, even you're working on a JS project. Like jsdocs but better.

evanshortiss:
Thanks @MikeyBurkman definitely need to review some of those.

abstractj:
@evanshortiss still want to move forward with this?

evanshortiss:
@abstractj yep, will take a look into these comments this evening.

keycloak.d.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * Perform live validation of an `access_token` against the Keycloak server.
+     *
+     * @param {Token|String} token The token to validate.
+     * @param {Function} callback Callback function if not using promises.
+     *
+     * @return {boolean} `false` if the token is invalid, or the same token if valid.
+     */
+    validateAccessToken(token: string): Promise<boolean>
```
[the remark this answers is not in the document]
evanshortiss: @MikeyBurkman yeah, I need to add the callbacks. Would like to just ignore, but I think that might be unreasonable!

Nicely spotted on the token front and thanks for the fix!

keycloak.d.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    /**
+     * Returns a user info JSON Object
+     * @param {String} token
+     */
+    userInfo(token: string): Promise<Object>
```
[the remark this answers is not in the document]
evanshortiss: Yeah, Object was a placeholder. I've found [this](https://www.keycloak.org/docs/3.0/securing_apps/topics/oidc/oidc-generic.html) that describes it as "standard claims". @abstractj, do you know if this object is defined someplace? Does it contain known properties, or are some dynamic?

I think it could be represented as something like this so application specific keys can be defined:

```ts
userInfo<T extends StandardClaims>(token: string): Promise<T>
```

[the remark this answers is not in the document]
evanshortiss:
```ts
userInfo<T extends Token|string, C extends StandardClaims>(token: T): Promise<C>
```

keycloak.d.ts
```diff
@@ -0,0 +1,348 @@
...
+     *
+     * This method creates the `Grant` object, including
+     * the `access_token`, `refresh_token` and `id_token`
+     * if available, and validates each for expiration and
+     * against the known public-key of the server.
+     *
+     * @param {String} rawData The raw JSON string received from the Keycloak server or from a client.
+     * @return {Promise} A promise reoslving a grant.
+     */
+    createGrant(data: GrantProperties): Promise<Grant>
```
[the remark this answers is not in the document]
evanshortiss: Ah, need to update docs. The function accepts the JSON string or Object.

keycloak.d.ts
```diff
@@ -0,0 +1,348 @@
...
+   * If any of the following errors are seen the promise will resolve with undefined:
+   *
+   * - The token was undefined in the first place.
+   * - The token is expired.
+   * - The token is not expired, but issued before the current *not before* timestamp.
+   * - The token signature does not verify against the known realm public-key.
+   *
+   * @return {Promise} That resolve a token
+   */
+    validateToken(token: Token): Promise<Token>
```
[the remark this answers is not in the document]
evanshortiss: It seems like we should simply enforce in the types that you must pass a token string? The JS currently has the following check which would reject with an error if the token is undefined, so the docs seem to be outdated anyway.

```ts
if (!token) {
  reject(new Error('invalid token (missing)'));
}
```

I'll update the docs to align with the current behaviour.

keycloak.d.ts
```diff
@@ -0,0 +1,351 @@
...
+     *
+     * @return {boolean} `false` if the token is invalid, or the same token if valid.
+     */
+    validateAccessToken<T extends Token|string>(token: T): Promise<false|T>
+
+    /**
+     * Returns a user info JSON Object
+     * @param {Token|String} token
+     */
+    userInfo<T extends Token|string, C extends StandardClaims>(token: T): Promise<C>
```
evanshortiss: WIP - We need to work out what `StandardClaims` needs to look like.

keycloak.d.ts
```diff
@@ -0,0 +1,351 @@
...
+     *
+     * This method accepts a token, and returns a promise
+     *
+     * If the token is valid the promise will be resolved with the token
+     * 
+     * If the token is undefined or fails validation an applicable error is returned
+     * 
+     * @return {Promise} That resolve a token
+     */
+    validateToken(token: Token): Promise<Token>
```
evanshortiss: Are the changes here ok? The previous comments (taken from the JS source) seemed outdated,

ssilvert: Looks fine code-wise, but @abstractj would have to say if the comments are correct.

BorntraegerMarc:
I would love to see this merged! I always find libraries with TS support so much easier to use :)

abstractj:
@BorntraegerMarc thank you. We're going to take this into consideration for future releases.

koldoon:
Our team would really appreciate this, it is _much_ easier to use libraries with typings.

keycloak.d.ts
```diff
@@ -0,0 +1,351 @@
...
+
+declare namespace KeycloakConnect {
+
+  interface KeycloakConfig {
+    'confidential-port': string|number
+    'auth-server-url': string
+    'resource': string
+    'ssl-required': string
+    'bearer-only'?: boolean
+    realm: string
```
svmn: Probably 
```
credentials: {
  secret: string;
};
```

should be added here.

abstractj:
@MikeyBurkman @evanshortiss @ssilvert @koldoon @svmn @BorntraegerMarc I started a thread on the mailing list so we can have a final decision on this http://lists.jboss.org/pipermail/keycloak-dev/2019-May/012029.html. Feel free to join and discuss with us.

evanshortiss:
@abstractj I just subscribed to that list. I assume I need to wait for the next reply to the mail thread to be able to send my 2 cents?

abstractj:
@evanshortiss I added you to the thread, I hope I send to the correct e-mail. Based on @ssilvert's comments, I believe we gonna plan it. If you know some way of automate (I'm not sure if it's technically possible) the creation of Typescript definitions, please feel free to reply on the same thread.

BorntraegerMarc:
Something like this could be used to generate JS Docs to typescript declaration files: https://github.com/englercj/tsd-jsdoc

BUT I think this PR is a very good first step to get to use typescript. In the end IMO this repo should only write typescript and transpile to JS

abstractj:
@ssilvert @evanshortiss I'm closing this because it was superseded by https://github.com/keycloak/keycloak-nodejs-connect/pull/204. What I just was just to rebase with the latest changes from our codebase.

@evanshortiss if there's anything which you would like to comment on that. Please do.

Kinrany:
Have you ever figured out the type of `StandardClaims` returned by `userInfo`?

PR #124, opened 2018-04-27
Add configurable host and port for Docker Compose

bdmayes:
Using keycloak-connect within a Docker service is currently tricky. The
module uses the same host and port for both the auth server as well as
the app server. However, with docker-compose it is extremely easy to
have the two running in separate containers, causing issues with what
exactly does localhost refer to, and also not properly mapping back to
the appropriate port number in the other container. This commit simply
adds the ability to specify these such that when a grant is obtained
from code, the host and port to connect to inside the http request
can be customized.

### DIFF
--- a/middleware/auth-utils/grant-manager.js
+++ b/middleware/auth-utils/grant-manager.js
@@ -39,6 +39,8 @@ function GrantManager (config) {
   this.public = config.public;
   this.bearerOnly = config.bearerOnly;
   this.notBefore = 0;
+  this.authServerHost = config.authServerHost;
+  this.authServerPort = config.authServerPort;
   this.rotation = new Rotation(config);
 }
 
@@ -99,6 +101,14 @@ GrantManager.prototype.obtainFromCode = function obtainFromCode (request, code, 
   };
   const handler = createHandler(this);
   const options = postOptions(this);
+
+  if (this.authServerHost) {
+    options.hostname = this.authServerHost;
+  }
+
+  if (this.authServerPort) {
+    options.port = this.authServerPort;
+  }
 
   return nodeify(fetch(this, handler, options, params), callback);
 };
--- a/middleware/auth-utils/config.js
+++ b/middleware/auth-utils/config.js
@@ -161,6 +161,16 @@ Config.prototype.configure = function configure (config) {
     }
     this.publicKey += '-----END PUBLIC KEY-----\n';
   }
+
+  /**
+   * Configurable host for keycloak auth server.
+   */
+  this.authServerHost = config['auth-server-host'] || config.authServerHost;
+
+  /**
+   * Configurable port number for keycloak auth server.
+   */
+  this.authServerPort = config['auth-server-port'] || config.authServerPort;
 };
 
 module.exports = Config;
--- a/test/fixtures/auth-utils/keycloak-docker.json
+++ b/test/fixtures/auth-utils/keycloak-docker.json
@@ -1,0 +1,10 @@
+{
+  "server-url" : "https://localhost:8080/auth",
+  "realm" : "nodejs-test",
+  "min-time-between-jwks-requests" : 0,
+  "ssl-required" : "external",
+  "resource" : "public-client",
+  "public-client" : true,
+  "auth-server-host" : "someFakeHost",
+  "auth-server-port" : 9090
+}
--- a/test/grant-manager-spec.js
+++ b/test/grant-manager-spec.js
@@ -536,3 +536,16 @@ test('GrantManager#validateToken returns undefined for an invalid token', (t) =>
   }
   t.end();
 });
+
+test('GrantManager#obtainFromCode should try to connect to a configurable host and port', (t) => {
+  const dockerManager = getManager('./test/fixtures/auth-utils/keycloak-docker.json');
+  const fakeRequest = {};
+  const fakeCode = 'a-fake-code';
+
+  dockerManager.obtainFromCode(fakeRequest, fakeCode)
+    .catch((err) => {
+      t.true(err instanceof Error, err.message);
+      t.equal(err.message, 'getaddrinfo ENOTFOUND somefakehost somefakehost:9090');
+    })
+    .then(t.end);
+});


middleware/auth-utils/grant-manager.js
```diff
@@ -100,6 +102,14 @@ GrantManager.prototype.obtainFromCode = function obtainFromCode (request, code,
   const handler = createHandler(this);
   const options = postOptions(this);
 
+  if (this.authServerHost) {
+    options.hostname = this.authServerHost;
+  }
+
+  if (this.authServerPort) {
+    options.port = this.authServerPort;
+  }
```
bdmayes: I'm a little unclear if these settings should only be set here, or if they should additionally be set inside of `obtainDirectly`, `obtainFromClientCredentials`, etc. If so, then the change should probably occur inside of the `postOptions` function, and the code would instead be the follow. I'm just not sure what the ramifications of that would be for other credential methods.

```
if (manager.authServerHost) {
  opts.hostname = manager.authServerHost;
}

if (manager.authServerPort) {
  opts.port = manager.authServerPort;
}
```

abstractj:
@bdmayes please file a Jira for this and attach this PR

bdmayes:
@abstractj I already opened the JIRA ticket a few days ago (hence the branch name here). I just added a link to this PR as a comment. I guess that is where you wanted it? In any case, see https://issues.jboss.org/projects/KEYCLOAK/issues/KEYCLOAK-7258?filter=allissues

till:
@bdmayes We ran into this as well — also running in Docker. Is there another better fix we are not aware of?

till:
@bdmayes Thanks for responding and also thanks for the code on here.

Btw, if people don't want to run on a "patched" branch, our work around was to add a dns entry `127.0.0.1 keycloak` to `/etc/hosts`. Then you configure `keycloak.json` with `http://keycloak:8080/auth` and it works _inside_ compose and on the client-side.

In prod, we'll use DNS as well. But it would be great if this can be merged. (cc @abstractj).

bdmayes:
Hey folks. It has been a while since I opened this PR. Any chance someone can review it and merge it if it looks good?

till:
@abstractj Someone already took the time to send you code. Maybe you can get them to be a maintainer down the road to help with these things. ;)

abstractj:
@till I understand, feel free to discuss this with @stianst. But we do have other tasks that have to be done first. If you are in a rush, the code is open source, which means that you can change to your own needs.

stianst:
I'm not convinced about this PR at all. auth-server-url is already configurable and if you want to specify it with two separate environment variables all you should need to do in keycloak.json is use:
    auth-server-url: ${env.KC_HOST}:${env.KC_PORT}

abstractj:
@bdmayes @till I'm closing because there's an alternative solution like @stianst mentioned. Please,  reopen the Jira, if the issue still persists. Thank you.

bdmayes:
Hi folks,

Sorry, I am just now getting back around to looking at this. I believe there is still an issue here, specifically for docker-compose environments which are running on localhost. As I mentioned before, the problem stems from the fact that the auth-server-url is used for two different purposes. It seems to be a callback URL on the app server, but it is also the auth URL. So let's take the following docker-compose.yml file as an example. You can see there are 3 services, app (a simple express app using keycloak-connect), a postgres DB, and keycloak. Notice that keycloak and app are both connected to the "keycloak" network. Furthermore, app exposes port 3000 out to the host, while keycloak exposes 8080 in the container out as 8083 on the host:

```
version: '3.4'

networks:
  keycloak:
  postgres:

services:

  app:
    image: node
    volumes:
      - .:/opt/app
    working_dir: /opt/app
    ports:
      - 3000:3000
    user: node
    environment:
      LOG_LEVEL: debug
      NODE_ENV: development
    networks:
      - default
      - keycloak
    command: npm run dev

  postgres:
    image: postgres
    volumes:
      - .:/var/lib/postgresql
    environment:
      POSTGRES_DB: keycloak
      POSTGRES_USER: keycloak
      POSTGRES_PASSWORD: keycloak
    networks:
      - postgres

  keycloak:
    image: jboss/keycloak:4.0.0.Final
    environment:
      DB_VENDOR: POSTGRES
      DB_ADDR: postgres
      DB_PORT: 5432
      DB_DATABASE: keycloak
      DB_USER: keycloak
      DB_PASSWORD: keycloak
      KEYCLOAK_USER: keycloak
      KEYCLOAK_PASSWORD: keycloak
      PROXY_ADDRESS_FORWARDING: "true"
    networks:
      - keycloak
      - postgres
    ports:
      - 8083:8080
    depends_on:
      - postgres
```

The question becomes, what value am I supposed to put into the keycloak config to get this to work? I started out by trying this line:

`"auth-server-url": "http://localhost:8083/auth",`

In that configuration, when I visit localhost:3000 in host browser, it will redirect me to localhost:8083. However, that gives me an error saying "We're sorry... page not found". I believe this is because once it hits my express API, it is looking at the auth-server-url in the grant manager trying to hit "localhost" on port 8083. However, the app server is running inside of a software defined network (SDN) setup by Docker, so when in the Express route, localhost is the app server where express is running, not the keycloak container that is actually serving on port 8080.

Similarly, I cannot simply change my config to point auth-server-url at "http://keycloak:8080/auth" since that is only accessible inside of the SDN created by Docker. That will cause localhost:3000 on the host system to 302 redirect to keycloak:8080, which is only accessible inside of the SDN and thus the page immediately fails to load anything.

Perhaps my variable names were not specific enough, leading to some confusion, or perhaps I am still missing something. I really think there is a problem with this variable being used to represent both the auth server as as well as the app server. When configured inside of a compose network, the term "localhost" depends on exactly which container you are in at any given point in time.

I am happy to be wrong, but kindly review and see what you think. If there is some other way to make this work and I am simply not configuring it correctly, that would be wonderful. At the moment however, I cannot make this work with the changes I posted in my PR.

stianst:
auth-server-url should point to your Keycloak server. It is not the callback URI for the Node.js application. Not sure how the Node.js adapter decides the callback URI, but it's definitively not based on auth-server-url. If it is that's just plain wrong. It also shouldn't have to be configured as it should in most cases be able to read it directly from the request URL.

abstractj:
@stianst I would like to move this discussion to https://issues.jboss.org/browse/KEYCLOAK-7258 because this PR is closed and it's easy to miss it. I will copy what you said there.

PR #126, opened 2018-05-14
Pass `next` to `accessDenied` method to allow error handling

lennym:
The `next` callback was not being passed in this instance of `keycloak.accessDenied` (as opposed to the [similar call on L27](https://github.com/lennym/keycloak-nodejs-connect/blob/42bff7edeb596548ed8c13d28e2389dba8c6b21b/middleware/post-auth.js#L27)). This means that implementations cannot easily handle the error using standard express error handling mechanisms.

### DIFF
--- a/middleware/post-auth.js
+++ b/middleware/post-auth.js
@@ -49,7 +49,7 @@ module.exports = function (keycloak) {
         }
         response.redirect(cleanUrl);
       }).catch((err) => {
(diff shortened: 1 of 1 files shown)


abstractj:
PR rebased and superseded by https://github.com/keycloak/keycloak-nodejs-connect/pull/149

PR #138, opened 2018-06-25
KEYCLOAK-7711 Remove nsp from product build pom

ASzc:
Missed this in the fix for KEYCLOAK-7684, same thing but the pom. I've actually
tried a local build this time, and it works, so this should be the last fix.


ASzc:
Duplicate for REL branch, see #137

PR #139, opened 2018-06-29
[KEYCLOAK-7738] Not validate expire date for offline refresh_token

lodaleaf:


abstractj:
@lodaleaf please make sure you squash your commits into a single one. Also, for each Jira, for example 
KEYCLOAK-7711 and KEYCLOAK-7738 split them in separate pull-requests.

lodaleaf:
Thanks for reviewing,I have fixed it

abstractj:
@lodaleaf do you have the code to reproduce the issue? I could not succeed trying to reproduce this here.

abstractj:
@lodaleaf thanks so much for spending time on this. But I could not reproduce the issue you mentioned using the latest releases. If the issue still persists, please, reopen the Jira adding the detailed steps or the code to reproduce the bug.

dschwank:
Currently I am facing the same issue. This fix would solve the problem.

**How to reproduce**
1. Request refresh and offline token (`offline_access`)
2. => Access token & Refresh token are saved in the store
3. Access token expires -> Request new access token using the refresh token
4. New access token is returned by keycloak, but this time keycloak returns an offline token
5. Access token expires -> ERROR

As an offline token has no `exp` time, the `isExpired` check will always be `true`. We solved the issue right now by disabling the offline token, but I would vote for including this solution!

@abstractj & @lodaleaf, can we merge this solution?

Let me know if I can support somehow!

PR #147, opened 2018-08-22
[KEYCLOAK-8116] Change Travis job from Node.js adapter to use Docker

abstractj:


abstractj:
@stianst before docker image build took 21 min to finish. Now 4 min :)

Huge win!

scripts/start-server.sh
```diff
@@ -1,37 +1,12 @@
...
-# Start the server
-$KEYCLOAK/bin/standalone.sh -Djava.net.preferIPv4Stack=true \
-                            -Dkeycloak.migration.action=import \
-                            -Dkeycloak.migration.provider=singleFile \
-                            -Dkeycloak.migration.file=./test/fixtures/auth-utils/nodejs-test-realm.json \
-                            -Dkeycloak.migration.strategy=OVERWRITE_EXISTING > keycloak.log 2>&1 &
-
-waitForServer
+# For debugging purposes to make sure the image was started
+docker ps -a
```
stianst: Perhaps it would be better to add a block tries to fetch a URL from Keycloak to make sure it's up? It should do it in a loop for X seconds until it's up.

abstractj: Humm great idea.

abstractj:
@stianst updated!

PR #153, opened 2018-09-05
KEYCLOAK-8241 Switch dependencySource to BOMREST in prod-arguments

ASzc:
This will allow dynamic alignment to versions suggested by DA


abstractj:
@ASzc could you please rebase?

PR #159, opened 2018-10-22
Permissions made available to other mw

ulrik-s:
Permissions fetched from the server by the enforcer might want to be
used by other middleware in express for futher processing e.g
filtering.


stianst:
This is missing JIRA and tests

pedroigor:
@ulrik-s, I've created https://issues.jboss.org/browse/KEYCLOAK-8640. Could you please change your commit to reference this JIRA and also provide tests ?

Otherwise, changes LGTM.

PR #160, opened 2018-10-23
[KEYCLOAK-8640] Expose granted permissions to other middleware

ulrik-s:
Permissions fetched from the server by the enforcer might want to be
used by other middleware in express for futher processing e.g
filtering.


pedroigor:
@ulrik-s, thanks.

PR #173, opened 2018-12-10
KEYCLOAK-9051: added request instance to support custom agents

tuqire:


abstractj:
@tuqire thanks for all the details you provided on Jira, it really helps. I believe would make sense to add some integration tests for this change. Maybe some test cases for the custom agent.

tuqire:
> @tuqire thanks for all the details you provided on Jira, it really helps. I believe would make sense to add some integration tests for this change. Maybe some test cases for the custom agent.

Hi,, that's no problem, thanks for the quick feedback.

Will add the integration tests and update.

tuqire:
@abstractj hi, I have added an integration test but tbh I am not sure if that is sufficient. Let me know if there is anything else specifically I should be testing.

PR #175, opened 2018-12-13
[KEYCLOAK-9047]: add keycloak.protect function support AsyncFunction

goldcome:
if protect function guard is AsyncFunction, guard  function will return promise, `if (!guard || guard(request.kauth.grant.access_token, request, response))` return true always , not what we expected, add support AsyncFunction

```
let kylib = {
    authenticate: async(token, req, res) => {
        try {
            xxxxxx            
        } catch(e) {
            console.log(e);
            return false;
        }
    }
};
module.exports = kylib;
```

```
const kylib = require("./kylib");
app.get('/login', keycloak.protect(kylib.authenticate), function (req, res) {
 
});

### DIFF
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -50,17 +50,31 @@ module.exports = function (keycloak, spec) {
 
   return function protect (request, response, next) {
     if (request.kauth && request.kauth.grant) {
-      if (!guard || guard(request.kauth.grant.access_token, request, response)) {
+      if (!guard) {
         return next();
+      } else {
+        if (guard.constructor.name === 'AsyncFunction') {
+          guard(request.kauth.grant.access_token, request, response).then(function (result) {
+            if (result) {
+              return next();
+            } else {
+              return keycloak.accessDenied(request, response, next);
+            }
+          });
+        } else {
+          if (guard(request.kauth.grant.access_token, request, response)) {
+            return next();
+          } else {
+            return keycloak.accessDenied(request, response, next);
+          }
+        }
       }
-
-      return keycloak.accessDenied(request, response, next);
-    }
-
-    if (keycloak.redirectToLogin(request)) {
-      forceLogin(keycloak, request, response);
     } else {
-      return keycloak.accessDenied(request, response, next);
+      if (keycloak.redirectToLogin(request)) {
+        forceLogin(keycloak, request, response);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: middleware/protect.js [hunk 1]
```diff
@@ -50,17 +50,31 @@ module.exports = function (keycloak, spec) {
...
   return function protect (request, response, next) {
     if (request.kauth && request.kauth.grant) {
-      if (!guard || guard(request.kauth.grant.access_token, request, response)) {
+      if (!guard) {
         return next();
+      } else {
+        if (guard.constructor.name === 'AsyncFunction') {
+          guard(request.kauth.grant.access_token, request, response).then(function (result) {
+            if (result) {
+              return next();
```
mhajas: I am not sure we need return value of `next()`, but this way `protect()` function doesn't return anything. Is it problem @abstractj @vmuzikar?

vmuzikar:
@abstractj I'm a bit confused. This is more than a year old PR and the JIRA says it's out of date. So is this still valid?

abstractj:
@vmuzikar it was closed by accident, reopened now.

middleware/protect.js
vmuzikar: @mhajas You mean that in case of async function we don't need to return anything? I think we still need to return `next()`, why wouldn't we?

mhajas: No I am not saying we don't need to return anything, but that we probably don't return anything when it is written as it is now. I am not sure how it works exactly with promises, but if I am not mistaken, this `return next()` doesn't belong to `protect()` function but to promise's `then()` function.

vmuzikar: @mhajas It seems `next()` in the `then` block really belongs to `protect`. Anyway, that's what my IDE tells me. :D

mhajas: I tried something like this:
```javascript
function a() {
  var promise = new Promise(function(resolve, reject) {
      resolve("Stuff worked!");
  });

  promise.then(function(result) {
  	console.log("then called")
    return "return from then";
  });
}

console.log("result is: " + a())
```

and it prints 

```
result is: undefined
then called
```

abstractj:
Closing it as out of date @goldcome if in the future you decide to update the PR with tests. Please, reopen and let us know.

PR #177, opened 2018-12-20
Use https always when constructing redirect_uri parameter

beatfactor:


beatfactor:
But the way the protocol is computed doesn’t really make it possible to choose in many cases. If the keycloak-proxy runs on http behind an nginx load balancer, for instance, the protocol will automatically be set to http without the possibility to be changed.

I don’t think that setting it to https always is necessarily a bad thing as it will ultimately force people to use https. They might not like it perhaps and it will probably be a bit more inconvenient, but it would lead to more secure websites. Giving the users an option to choose a less secure option for their convenience is not really helping anyone.

abstractj:
@beatfactor I understand and I'm not saying it's a bad thing. But we need tests to pass and flexibility for people. Let's move this discussion the mailing list.

PR #181, opened 2019-01-17
redirect() api requires a callback

winster:
Server getting crashed, when directly accessing an API with middleware Keycloak.protect(). The reason being, upon redirect a next function is expected

### DIFF
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -17,7 +17,7 @@
 
 const UUID = require('./../uuid');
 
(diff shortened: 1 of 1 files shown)


abstractj:
@winster for any change we require not only code, but also tests and a proper Jira containing all the details. Anything that does not include these requirements is considered incomplete and cannot be merged. 
If you are able to add the missing parts, please do so and feel free to reopen. Until then, we must close this issue.

PR #186, opened 2019-03-13
Redirect to register page in case of register

OmerWaleed:
We have developed a new method to build a register route. Protect function takes a second parameter i.e. register and decides to build a register url same as login url does.


abstractj:
@OmerWaleed for all features we require not only code, but tests, Jira, documentation and a discussion on keycloak-dev mailing list. Any submission that does not include all three is incomplete and cannot be merged. If you are able to add the missing parts, please do so and we can reopen and reconsider. Until then, we must close this issue.

PR #187, opened 2019-03-22
KEYCLOAK-9820

abstractj:
@hmlnarik @stianst after your approval on this, I'm going to update the prod repo. Here some considerations:

* There were two dependencies which required some upgrades Semistandard and Istanbul. `Semistandard` is used to scan for bad practices into the code and also code formatting, while `Istanbul` is used for code coverage.
* After the upgrade `Semistandard` reported several bad smells into our codebase, including formatting and error handling. Both were fixed. That's the reason why this PR touches 14 files.
* Istanbul was deprecated and moved to `nyc` project
* For semistandard there are still some issues reported with npm audit. To make sure that the versions are locked and we're using the most up to date dependencies a new `package-lock.json` file was generated.
* You can see that there are no issues anymore by running `npm audit` 

If there are any questions, please let me know.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,155 @@
 {
   "name": "keycloak-connect",
-  "version": "4.8.2-dev",
+  "version": "6.0.0-dev",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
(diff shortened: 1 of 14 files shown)


package-lock.json
```diff
@@ -2274,7 +2274,7 @@
       "dependencies": {
         "es6-promise": {
           "version": "3.0.2",
-          "resolved": "https://registry.npmjs.org/es6-promise/-/es6-promise-3.0.2.tgz",
+          "resolved": "http://registry.npmjs.org/es6-promise/-/es6-promise-3.0.2.tgz",
```
hmlnarik: Why is there the change from https to http? (same with readable-stream dep)

abstractj: @hmlnarik because that's the standard in new versions of npm. Do you see any blockers?

stianst: Strange. So while the rest of the world is going to https always NPM is for some reason going backwards?

hmlnarik: No blockers, the switch from https to unsecured http just seemed a typo to me (particularly with almost of the other dependencies still using https)

abstractj: @hmlnarik @stianst let me rephrase myself. Not a standard, but known bug https://github.com/npm/npm/issues/20719

abstractj:
@hmlnarik it's not bad, it's awful. Until npm fix it, we can leave it as is.

PR #188, opened 2019-04-11
KEYCLOAK-10035 Align token error message with Java adapters

mhajas:
I am not sure whether you require keycloak issue for each PR in nodejs adapter. If yes let me know I will create one.


stianst:
All changes should have a JIRA so please create one

abstractj:
@mhajas just amend to your commit message, please. Something like [KEYCLOAK-10035] Align token error message with Java adapters.

@mhajas one thing that would be great is to add tests to check for these error messages. So we make sure that each condition is covered. Does it make sense?

mhajas:
@abstractj sure, no problem. However, I have quite a lot of work to do and this doesn't have a big priority, so I added it to my TODO list. Hopefully, I will find time for it next week.

PR #194, opened 2019-04-18
[KEYCLOAK-9683] Add contributors guidelines to keycloak-nodesjs-connect

abstractj:


abstractj:
@vmuzikar you were correct. It should be updated now.

PR #196, opened 2019-05-15
KEYCLOAK-8359 Migrate arquillian tests to nodejs testsuite

mhajas:


package.json
```diff
@@ -50,6 +50,7 @@
     "keycloak-request-token": "^0.1.0",
     "nock": "^9.0.2",
     "phantomjs-prebuilt": "^2.1.14",
+    "chromedriver": "^2.41.0",
```
mhajas: I left this here only for testing purposes, this should be probably removed when this PR is ready to merge.

test/keycloak-connect-web-spec.js
```diff
@@ -71,6 +73,62 @@ test('Should login with admin credentials', t => {
...
+  page.login('test-admin', 'password');
+
+  return page.events().getText().then(text => {
+    t.equal(text, 'Auth Success', 'User should be authenticated');
+    return page.output().getText().then(firstToken => {
+      page.logInButton().click(); // Invoke login for the second time, token shouldn't be changed
+      return page.output().getText().then(secondToken => {
+        t.equal(secondToken, firstToken, 'Token should not be changed as first session is still valid');
+        page.logOutButton().click();
+        return page.output().getText().then(text => {
```
mhajas: This test is not working for me in phantomjs. Me and @vmuzikar tried many things to make it work, however we were not successful. In chrome it works well.

test/utils/webdriver.js
```diff
@@ -19,9 +19,24 @@
...
+// o.addArguments('start-fullscreen');
+o.addArguments('disable-infobars');
+// o.addArguments('headless'); // running test on visual chrome browser
+o.setUserPreferences({ credential_enable_service: false });
+// const driver = new webdriver.Builder()
+//         .setChromeOptions(o)
+//         .forBrowser('chrome')
+//         .build();
+
 const driver = new webdriver.Builder()
```
mhajas: To run tests in chrome just comment this variable definition and uncomment the on above.

test/keycloak-connect-web-spec.js
abstractj: @mhajas could you please provide your error output? Just to make sure we're talking about the same thing.

package.json
abstractj: Let's just remove it, so we can focus on a single issue.

test/keycloak-connect-web-spec.js
mhajas: @abstractj Here is the output of my run: https://pastebin.com/smz7ECaT

mhajas:
@abstractj Thanks for looking into it. lint should be ok now.

mhajas:
This PR need to be merged after https://github.com/keycloak/keycloak-nodejs-connect/pull/198

mhajas:
@abstractj @vmuzikar I little bit struggled with making Travis green, there was some interference between test files (token-test file was failing when running after grant-manager-spec.js file). I changed travis to run each test file separately as we are doing everywhere else. If we want to run files using tape in two batches as it was before, I will probably need some discussion/help from you guys.

mhajas:
@abstractj Here is the issue we had discussion about. In current master tests are running correctly with `tape`, however for this PR I am able to make tests run only in case I run them in for loop. I changed `run-tests.sh` to use tape so that we can see whether Travis fails.

abstractj:
@mhajas this should fix your issue with `nock` https://github.com/abstractj/keycloak-nodejs-connect/commit/ce8e6f5a1dcd46e1eefa32ed4cf58e6a3a9028aa and unblock you to the next level. Regards whether or not we should bring back the for loop into run-tests.sh, I believe the answer is yes, as long as we check for the exit codes.

mhajas:
@abstractj thank you very much for your fix. I updated run-tests.sh to use loop and check for error codes (I added also some additional logging since nodejs runs are not very talkative) and also cherry-picked you commit into this PR.

PR #199, opened 2019-05-24
Added baseUrl for all redirects while post-auth

DrBoria:
Fixes for redirecting to baseUrl instead of '/'


abstractj:
@DrBoria thanks for submitting this change. but please read https://github.com/keycloak/keycloak/blob/master/CONTRIBUTING.md and follow these guidelines when contributing to Keycloak. After you create a Jira and write tests, we can start to review this change.

PR #204, opened 2019-06-20
[KEYCLOAK-6409] TypeScript typings for keycloak-nodejs-connect

abstractj:
@ssilvert I had to rebase and update that PR. Could you please review when you get the chance?


evanshortiss:
@abstractj no objections here. Though I'm not sure if some internal keycloak interfaces/classes have changed since I made the PR. It's a good start though, and can always be iterated upon to improve 👍

keycloak.d.ts
```diff
@@ -0,0 +1,351 @@
...
+     *
+     * @return {boolean} `false` if the token is invalid, or the same token if valid.
+     */
+    validateAccessToken<T extends Token|string>(token: T): Promise<false|T>
+
+    /**
+     * Returns a user info JSON Object
+     * @param {Token|String} token
+     */
+    userInfo<T extends Token|string, C extends StandardClaims>(token: T): Promise<C>
```
korommatyi: Sorry to comment on a closed PR, but this line, more specifically the mention of `StandardClaims` prevents me to compile this. As far as I see, `StandardClaims` is not defined anywhere.

ssilvert: If you just remove the "extends StandardClaims", is that good enough for you?

ssilvert: If so, we can just do a quick fix for it.

MikeyBurkman: I just want to point out that if you just remove `extends StandardClaims`, then you're basically saying "this will return whatever the caller wants it to return" and that's probably less than ideal. Maybe good for a quick fix but I wouldn't leave it in the long term.

For the record, there was a TODO in the original PR that got lost when it was turned into this one https://github.com/keycloak/keycloak-nodejs-connect/pull/123/files#r194166591

ssilvert: Yes.  I don't have time to work out what StandardClaims needs to be.  If someone in the community (you?) wants to improve this file, please let me know.

So as long as my earlier suggestion will suffice for now, I can do a quick PR.  Have you tried it out just removing "extends StandardClaims"?

korommatyi: > If you just remove the "extends StandardClaims", is that good enough for you?

Yep, then it compiles without any error.

ssilvert: OK.  Preliminary fix is now merged: https://github.com/keycloak/keycloak-nodejs-connect/pull/210

korommatyi: Great, thanks!

PR #205, opened 2019-07-01
KEYCLOAK-10187 JUnit report style

slaskawi:
https://issues.jboss.org/browse/KEYCLOAK-10187

This commit introduces xUnit style reports for TAP (see the [project page](https://github.com/aghassemi/tap-xunit)). This should allow us to report failures in Jenkins UI.

Still to be done:

* [x] Make Travis green.
* [x] Prepare configuration for Jenkins and attach a screenshot to JIRA.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.tar
 *.tmp
 node_modules/
+test_results/
 /keycloak-*/
 coverage
 pid.txt
--- a/package-lock.json
+++ b/package-lock.json
@@ -1057,6 +1057,12 @@
         "domelementtype": "1"
       }
     },
+    "duplexer": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
+      "integrity": "sha1-rOb/gIwc5mtX0ev5eXessCM0z8E=",
+      "dev": true
+    },
     "ecc-jsbn": {
       "version": "0.1.2",
       "resolved": "https://registry.npmjs.org/ecc-jsbn/-/ecc-jsbn-0.1.2.tgz",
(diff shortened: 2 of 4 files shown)


slaskawi:
Please look at the JIRA for an example test report.

slaskawi:
> Would it make sense to add the test_results directory to .gitignore similarly to e.g. node_modules?

@vmuzikar That's a very good idea. Let me add it.

> Jenkins doesn't "group" some of the test files properly. I'm not sure why though - don't see anything wrong with the XML files.

I've looked at the [Surefire XSD file](https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report-3.0.xsd) to understand what is the expected format of a test execution. It seems we have 1 `testsuite` with a bunch `testcase`s. [Tap-xunit](https://github.com/aghassemi/tap-xunit) seems to violate this and seems to be using multiple `testsuite`s:

```xml
<?xml version="1.0"?>
<testsuites>
  <testsuite tests="1" failures="0" errors="0" name="abc.# GrantManager with empty configuration">
    <testcase name="#1 should throw"/>
  </testsuite>
  <testsuite tests="4" failures="0" errors="0" name="abc.# GrantManager with rogue configuration">
    <testcase name="#2 should be equal"/>
    <testcase name="#3 should be equal"/>
    <testcase name="#4 should be equal"/>
    <testcase name="#5 should be equal"/>
  </testsuite>
  <testsuite tests="1" failures="0" errors="0" name="abc.# GrantManager in public mode should be able to obtain a grant">
    <testcase name="#6 should not be equal"/>
  </testsuite>
...
```

I guess, that's where the aggregation problem comes from. I think we have two choices - we can either accept with this inconvenience (what's worth to mention - failures are interpreted correctly, and that matters the most) or we can try some other TAP plugins.

PR #207, opened 2019-07-02
KEYCLOAK-9996 Introduction of signature utility

abstractj:
- Prepare to support other algorithms in the future like HS256
- Migrate from roi to axios
- Update all the integration tests
- Add testing script to prevent concurrency issues


middleware/auth-utils/signature.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Verify signed data using the token information provided
+ * @TODO in the future provide more alternatives like HS256 support
+ * @param {Token} the Token object
+ */
+Signature.prototype.verify = function verify (token, callback) {
+  return new Promise((resolve, reject) => {
+    const verify = crypto.createVerify('RSA-SHA256');
```
pedroigor: Don't you need to consider the `alg` within the token header?

abstractj: For this initial PR, it's going to be hard-coded as is and this will be the only `alg` supported.

In the future, we may take others into consideration.

abstractj:
> Overall looks good but I'd rather leave @mhajas / @vmuzikar to step in. Few items:
> 
> * Is there a JIRA for this PR?

Yes, is the same id of the branch. But was my failure to not add this to my commit message.

> * Should the new `Error`s start in signature.js with a capital letter?

I don't have a strong opinion about it. But for now I'd rather to keep `prod` and `upstream` with similar error messages. Nothing stops us from update all the messages if we think it's reasonable.

middleware/admin.js
```diff
@@ -35,50 +38,63 @@ function adminLogout (request, response, keycloak) {
...
-        return;
-      }
-      if (sessionIDs && sessionIDs.length > 0) {
-        let seen = 0;
-        sessionIDs.forEach(id => {
-          keycloak.unstoreGrant(id);
-          ++seen;
-          if (seen === sessionIDs.length) {
+      signature = new Signature(keycloak.config);
+      signature.verify(token).then(token => {
```
mhajas: @abstractj It looks like we are not checking resource name. Is it correct?

abstractj: @mhajas I believe we handle this into the Grant Manager. Anyway, this PR is just a cherry-pick from what we already merged in the prod. So my suggestion is for us to get it merged and file a bug if there's anything missing.

@mhajas @hmlnarik that works for you?

mhajas: Hello, I tried to add this test:

````javascript
test('Test resource name', t => {
  t.plan(1);

  var app = new NodeApp();
  var client = admin.createClient(app.confidential('adminapp8'), realmName);

  return client.then((installation) => {
    app.build(installation);
    let opt = {
      method: 'post',
      url: `${app.address}/k_logout`,
      data: TestVector.logoutValidPayload
    };
    return axios(opt).then(response => {
      t.equal(response.status, 200);
    }).catch(err => {
      t.fail(err.response.data);
    });
  }).then(() => {
    app.destroy();
  });
});
````

I created the test like this: I took the valid scenario and changed the resource name of the deployment to something different (adminapp3 -> adminapp8). Should it be possible to logout from adminapp8 with token issued for resource adminapp3?

File: test/utils/helper.js [hunk 3]
```diff
@@ -51,6 +51,12 @@ function TestVector () {
 }
 
 TestVector.wrongRealmPublicKey = 'MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAikGBvYGniAJ59ZjpaSDw2o+j40Ila/dWfN8qA1dzXJesH9Z1sZrcevJB+rfxoZDaWMz2l9Q3OxG/qolTpsQl8NBdb5tymic9qDkAIsiyKThzjcfs5lOSxfnkHn6+Z0QbrYnXQs/cGvQ1Ai81M1M1O6BHDWu05n8c977h+BsfLmqGj7MZZj9gw9RM84RIKDGHTbFh9YyXBJVtqbOhRD7hcB0O9olDZb7mQ5A8gsMctcUhsVBy3xKCLMD41XU92rQ9FAlsV9mBglLqaVWr2mxQItN3lgjE02L8UyorI3T0uprIsnv7B2NwUC5ZhwZGfnBznUPVrT6makEJklpg5if3qQIDAQAB';
+TestVector.logoutValidPayload = 'eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsImtpZCI6IkZKODZHY0YzalRiTkxPY280TnZaa1VDSVVtZllDcW9xdE9RZU1mYmhObEUifQ.eyJpYXQiOjE1NTY2MTcwNzksInJlc291cmNlIjoiYWRtaW5hcHAzIiwiYWN0aW9uIjoiTE9HT1VUIiwibm90QmVmb3JlIjoxNTg3MDQ3NTM3fQ.NplTHo8JuwtmUbpp3AHjM3c6rn7g_xGWegC-b8Gg7V2QoN9vPRb9oCc9fdD7qWKpXLgfNtvtTIJnIIP5O_ux7Jt_SQyNtwoPmf5k_EFmm7JSxPnVfVA36BJbGDJu_BiNbktGgpNVZR5HnAkawhsLLo05S0edFfnbs4N9a_4W_YM';
```
mhajas: If I understand this correctly, a test using this logoutValidPayload will start to fail on 16th April next year because of the notBefore policy set to value: 1587047537? Shouldn't we create tokens in some other way? For example, setting notBefore for each test run?

vmuzikar: @mhajas Good catch! ;) However, I believe the test will be still working (regardless `notBefore` field) - please take a look at:
https://github.com/keycloak/keycloak-nodejs-connect/blob/c6205b8f7ceca23a7a4e431f2ec829b3335e8170/test/keycloak-connect-rest-mixed-client-spec.js#L218-L222
We're just testing the logout endpoint response which will always be `200`.

When reviewing the original PR in prod, I've actually given `notBefore` in this token a lot of attention. :) And what we're doing here in the new tests is, we're not testing the logout functionality as such. We're just testing the token signature validation, regardless if the logout is actually performed. So from my PoV, the `logoutValidPayload` is ok.

mhajas: I see it now. Thank you for the explanation.

PR #216, opened 2019-08-07
[KEYCLOAK-11016] Node.js build on Travis is broken due to chromedriver updates

abstractj:
@hmlnarik we might need to backport this to prod if we still have time. This won't affect the way how the adapter works, but people will have issues by running integration tests.


vmuzikar:
In general looks good but we need to wait and see what Travis has to say.

PR #217, opened 2019-09-03
[KEYCLOAK-11266] Add ability to specify hostname in config

kosciak9:
https://issues.jboss.org/browse/KEYCLOAK-11266 issue described here. I believe that linting and testing passes. Let me know what you think.


stianst:
I'm no node.js expert, but I would assume request.hostname should refer to the correct hostname as long as the proxy is passing the correct headers?

kosciak9:
You're right, that should happen. I'm going to investigate on my end.

stianst:
Closing this for now. Feel free to re-open if still needed.

PR #221, opened 2019-11-04
KEYCLOAK-11885 Include typings in the build

wederer:


wederer:
There seems to be an issue with the build in Travis. I am pretty sure this has nothing to do with my change, because the build before mine also failed with the same error.

PR #223, opened 2019-11-08
[KEYCLOAK-11978] Integration tests failing for the Node.js adapter

abstractj:
## What
* Integration tests for the Node.js adapter are failing due to chromedriver. We need to upgrade its dependencies, in order to make the build pass again.

## Expected Results
* Builds green on Travis

## Actual Results
* Integration tests failing on Travis when pull-requests are submitted: https://travis-ci.org/keycloak/keycloak-nodejs-connect

## Steps to reproduce
1. Clone the Node.js adapter repository
2. Run `make`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -148,13 +148,13 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/@types/events/-/events-3.0.0.tgz",
       "integrity": "sha512-EaObqwIvayI5a8dCzhFrjKzVwKLxjoG9T6Ppd5CEo07LRKfQ8Yokw54r5+Wq7FaBQ+yXRvQAYPrHwya1/UFt9g==",
-      "dev": true
+      "optional": true
     },
     "@types/glob": {
       "version": "7.1.1",
       "resolved": "https://registry.npmjs.org/@types/glob/-/glob-7.1.1.tgz",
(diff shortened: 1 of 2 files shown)


abstractj:
@mhajas we could use something like `"chromedriver": "*"`, it's the only way I know. Wdyt?

vmuzikar:
> @abstractj Is there a way to automatically use the latest version of chromedriver for testing?

@mhajas It's a great idea (we do the same for Keycloak UI tests) but we'd need Chrome to automatically update as well in Travis and our RHEL images. The thing is, each version of chromedriver supports only pretty narrow selection of versions of Chrome, i.e. each chromedriver release drops support for older Chrome versions.

abstractj:
@vmuzikar @mhajas updated, please let me know wdyt.

vmuzikar:
@abstractj  Do we know how often is Chrome updated in Travis? As I mentioned, I'm worried that we might get a new chromedriver version at some point in the time that won't support Travis' Chrome.

abstractj:
> @abstractj Do we know how often is Chrome updated in Travis? As I mentioned, I'm worried that we might get a new chromedriver version at some point in the time that won't support Travis' Chrome.

It seems like they just follow Chrome releases (https://docs.travis-ci.com/user/chrome#selecting-a-chrome-version): 
> Take note that the underlying Chrome version can change from one build to another when Google updates the stable or beta version.

PR #208, opened 2019-07-10
fix: errors in typings

darahayes:
I've copied the `keycloak.d.ts` file directly into my TypeScript project while I wait for a new release that contains the types. However I've found a couple of minor issues which I've fixed in this PR.

PR #225, opened 2019-11-29
KEYCLOAK-12099 Fix errors in typings

wederer:
See Jira Issue: https://issues.jboss.org/browse/KEYCLOAK-12099


wederer:
I created this PR as a copy to #208, because it had the wrong name and was stale. However I just saw that it was assigned recently and will thus close this one.

PR #226, opened 2019-12-03
KEYCLOAK-12042 Fix doubled slash in realm urls

mhajas:

### DIFF
--- a/middleware/auth-utils/config.js
+++ b/middleware/auth-utils/config.js
@@ -122,7 +122,7 @@ Config.prototype.configure = function configure (config) {
    * Authentication server URL
    * @type {String}
    */
-  this.authServerUrl = resolveValue(config['auth-server-url'] || config['server-url'] || config.serverUrl || config.authServerUrl);
+  this.authServerUrl = (resolveValue(config['auth-server-url'] || config['server-url'] || config.serverUrl || config.authServerUrl) || '').replace(/\/$/, '');
 
   /**
    * Root realm URL.
--- a/scripts/start-server.sh
+++ b/scripts/start-server.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # Pull the docker image with the latest changes on Keycloak
-docker pull jboss/keycloak:master
+docker pull quay.io/keycloak/keycloak
 
 # Start the image with default user/admin and import a default realm for integration tests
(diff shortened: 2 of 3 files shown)


mhajas:
Successful run in pipeline: https://keycloak-jenkins.com/job/universal-test-pipeline-adapters/915/

scripts/start-server.sh
```diff
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # Pull the docker image with the latest changes on Keycloak
-docker pull jboss/keycloak:master
+docker pull jboss/keycloak:latest
```
abstractj: @mhajas because you touched this file, I have to ask. What do you think about using quay.io instead? In this way we keep consistency across our repos.

abstractj:
@mhajas overall looks good to me. Just one minor comment regards quay.io. Please let me know what do you think when you get the chance.

vmuzikar:
Thanks @mhajas!

There's one potential issue and that is backward compatibility. It won't be possible to use newer adapter with older server. I'm not sure but do we even support this scenario? Maybe we support it the other way around – older adapter with newer server. Correct? (And this scenario is broken anyway.)

scripts/start-server.sh
mhajas: Well I can't see any reason why not to use quay. I will try to add it and see whether Travis passes.

mhajas:
You are right @vmuzikar I was looking into other adapters and it looks like js adapter supports both with and without endind /. There is an if here: https://github.com/keycloak/keycloak/blob/master/adapters/oidc/js/src/main/resources/keycloak.js#L681

We can do the same thing here. WDYT @vmuzikar @abstractj

mhajas:
@abstractj @vmuzikar I fixed the issue with backward compatibility. Could you have another look please?

abstractj:
@mhajas I believe we need to fix the issues reported by `make lint`

middleware/auth-utils/config.js
```diff
@@ -122,7 +122,7 @@ Config.prototype.configure = function configure (config) {
    * Authentication server URL
    * @type {String}
    */
-  this.authServerUrl = resolveValue(config['auth-server-url'] || config['server-url'] || config.serverUrl || config.authServerUrl);
+  this.authServerUrl = (resolveValue(config['auth-server-url'] || config['server-url'] || config.serverUrl || config.authServerUrl) || "").replace(/\/$/, "");
```
vmuzikar: @mhajas Just a stupid question, but why's there ` || ""`? Do we expect `resolveValue` to return some falsy value?

mhajas: I was surprised too, but it may return undefined. It happens in this test: https://github.com/keycloak/keycloak-nodejs-connect/blob/master/test/grant-manager-spec.js#L18

mhajas:
> @mhajas I believe we need to fix the issues reported by `make lint`

@abstractj done.

PR #228, opened 2019-12-04
Better Realm Error Handling

kevinarthur2077:
Makes error message more detailed when access is denied due to a mismatching issuer.


c0bra:
This would have saved me a hell of a lot of time.

PR #230, opened 2020-01-15
[KEYCLOAK-12428] Extra '/' character in the client > OIDC Json installation file

abstractj:
## JIRA ID
[KEYCLOAK-12428](https://issues.redhat.com/browse/KEYCLOAK-12428)

## Additional Information
The Node.js adapter is broken when urls contain a forward slash by the end of the URL. For example: http://localhost:8080/auth///////

## Verification Steps

1. Follow the steps to reproduce described in the same Jira
2. Run make tests

### DIFF
--- a/middleware/auth-utils/config.js
+++ b/middleware/auth-utils/config.js
@@ -122,7 +122,7 @@ Config.prototype.configure = function configure (config) {
    * Authentication server URL
    * @type {String}
    */
-  this.authServerUrl = (resolveValue(config['auth-server-url'] || config['server-url'] || config.serverUrl || config.authServerUrl) || '').replace(/\/$/, '');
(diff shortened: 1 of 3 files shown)


abstractj:
Thanks for reviewing this @tanaceus

abstractj:
> Thank you @abstractj. The PR looks good, it fixes the linked Jira. Just one thing to think about. Is this a valid scenario we are trying to fix? I mean if someone provides such configuration with more than one slash at the end of the auth-url, isn't it a reason to fail? Is there any chance of having more slashes in a valid auth-url? Because after this fix, it won't be possible to use node.js adapter with such authorization server.

@mhajas Indeed the scenario you mentioned is unlikely to happen. The reason for that testing scenario is to make sure that our adapter won't crash under any situation. But honestly, I'm not picky about this and I can update from `http://localhost:8080/auth//////////` to `http://localhost:8080/auth/` if you think it's better.

Please, let me know what works best for you.

PR #234, opened 2020-02-10
KEYCLOAK-12951 Add run-tests.sh to RH-SSO dist

vmuzikar:

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
     "test",
     "index.js",
     "uuid.js",
-    "keycloak.d.ts"
+    "keycloak.d.ts",
+    "run-tests.sh"
   ],
   "typings": "keycloak.d.ts",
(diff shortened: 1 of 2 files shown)


vmuzikar:
@abstractj Thanks. It'd be awesome if we could merge it before 9.0.0 – well need that for testing (even though it's not entirely blocking) and 2 other PRs rely on it. ;)

abstractj:
@vmuzikar which PRs?

vmuzikar:
@abstractj A backport to prod repo (this after all concerns only RH-SSO) and a PR to update the test pipelines. Neither of them is created (yet). ;)

The purpose of this is to unify testing of Keycloak and RH-SSO and to remove some hardcoded stuff from pipelines that could cause some trouble in future RH-SSO testing.

PR #237, opened 2020-02-27
KEYCLOAK-12771 Added missing type definitions

kdankert:
Added missing type definitions of functions that are exposed by the keycloak-connect module

Unfortunately nobody replied in the issue tracker if I could get assigned for this issue, but I decided to publish this PR anyways, because I want to fix this issue before keycloak-connect v9.0.1 is released.

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -8,11 +8,11 @@ import * as express from 'express'
  * To achieve this we export "KeycloakConnect" that references a namespace
  * containing our typings, and a static instance exposing the constructor
  */
-declare const KeycloakConnect: KeycloakConnectStatic
+declare const KeycloakConnect: KeycloakConnectStatic;
 export = KeycloakConnect
 
 interface KeycloakConnectStatic {
-  new (options: KeycloakConnect.KeycloakOptions, config: KeycloakConnect.KeycloakConfig): KeycloakConnect.Keycloak
+  new (options?: KeycloakConnect.KeycloakOptions, config?: KeycloakConnect.KeycloakConfig|string): KeycloakConnect.Keycloak
 }
 
 declare namespace KeycloakConnect {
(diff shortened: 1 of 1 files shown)


bender316:
@kdankert  As mentioned in the Issue it would be great if two other fixes could be included in this PR:

access_token, refresh_token and id_token in interface GrantProperties are currently not defined as type Token but as strings (which is totally wrong):
```javascript
interface GrantProperties {
    access_token?: Token
    refresh_token?: Token
    id_token?: Token
    expires_in?: string
    token_type?: string
  }
```

And in GuardFn the access_token is passed as type Token not as string
```javascript
type GaurdFn = (accessToken: Token, req: express.Request, res: express.Response) => boolean
```

kdankert:
Is there the Option to include this in 9.0.1 as this doesn't require any features to be added. It's a non breaking change. Would really like to see this in production

keycloak.d.ts
```diff
@@ -192,7 +192,20 @@ declare namespace KeycloakConnect {
...
-  type GaurdFn = (accessToken: string, req: express.Request, res: express.Response) => boolean
+  type GaurdFn = (accessToken: Token, req: express.Request, res: express.Response) => boolean
+
+  interface EnforcerOptions {
+    response_mode?: string,
+    resource_server_id?: string,
+    claims?: (...args: any[]) => any
+  }
+
+  interface AuthZRequest {
```
vmuzikar, line 203: I think `claim_token_format` is missing here per:
https://github.com/keycloak/keycloak-nodejs-connect/blob/043807542cb94d5477da867808e9b5b559f9eddf/middleware/auth-utils/grant-manager.js#L123

kdankert: You are right. I added it and will push the commit in a few minutes. Any idea when this is going to be released?

gardlt:
LGTM. I agree with the implementation. Will this be included in the `9.0.*` release?

PR #245, opened 2020-04-20
[KEYCLOAK-13841] Update typescript declarations

ckwalsh:

PR #247, opened 2020-04-23
fix: spec should be an optional param for protect

evanshortiss:
This is a simple one. The spec parameter is optional, but the types incorrectly specify it as required.


evanshortiss:
Ignore this. It's fixed in #245

PR #250, opened 2020-06-24
[KEYCLOAK-14578] Update KeycloakConfig types

beevelop:
Some Types are missing in the current implementation.
Related Jira-Issue: https://issues.redhat.com/browse/KEYCLOAK-14578

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
 declare namespace KeycloakConnect {
 
   interface KeycloakConfig {
-    'confidential-port': string|number
     'auth-server-url': string
-    'resource': string
-    'ssl-required': string
+    'realm': string
+    'clientId': string
+    'secret': string
+    'resource'?: string
+    'public-client'?: string
+    'realmUrl'?: string
+    'realmAdminUrl'?: string
+    'minTimeBetweenJwksRequests'?: number
     'bearer-only'?: boolean
-    realm: string
+    'plainKey'?: string
+    'verifyTokenAudience'?: boolean
   }
 
   interface KeycloakOptions {


beevelop:
Looks like there are a few more changes in PR #245. So I'd say whatever works best for the reviewers. If the other changes are valid too, I can cherry-pick the respective commits into this PR.

abstractj:
@beevelop if you are ok about this idea, let's do this. We cherry-pick changes from #245 keeping the authorship and update this pull-request with your commit. Thank you!

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
   interface KeycloakConfig {
-    'confidential-port': string|number
     'auth-server-url': string
-    'resource': string
-    'ssl-required': string
+    'realm': string
+    'clientId': string
+    'secret': string
+    'resource'?: string
+    'public-client'?: string
```
bruegth, line 26: Type should be changed into boolean here: https://github.com/keycloak/keycloak-nodejs-connect/blob/74a0d7fd45a74dfb2554cdd5e920d0ef641c684b/middleware/auth-utils/config.js#L117

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
 declare namespace KeycloakConnect {
 
   interface KeycloakConfig {
-    'confidential-port': string|number
     'auth-server-url': string
-    'resource': string
-    'ssl-required': string
+    'realm': string
+    'clientId': string
```
bruegth, line 23:
```suggestion
    clientId: string
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'secret': string
+    'resource'?: string
```
bruegth, line 25:
```suggestion
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
 declare namespace KeycloakConnect {
 
   interface KeycloakConfig {
-    'confidential-port': string|number
     'auth-server-url': string
-    'resource': string
-    'ssl-required': string
+    'realm': string
```
bruegth, line 22:
```suggestion
    realm: string
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'clientId': string
+    'secret': string
```
bruegth, line 24:
```suggestion
    secret: string
```

keycloak.d.ts
bruegth: 'resource' or 'client-id' see here: https://github.com/keycloak/keycloak-nodejs-connect/blob/74a0d7fd45a74dfb2554cdd5e920d0ef641c684b/middleware/auth-utils/config.js#L107

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'resource'?: string
+    'public-client'?: string
+    'realmUrl'?: string
```
bruegth, line 27:
```suggestion
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'realmAdminUrl'?: string
```
bruegth, line 28:
```suggestion
```

keycloak.d.ts
bruegth: see here https://github.com/keycloak/keycloak-nodejs-connect/blob/74a0d7fd45a74dfb2554cdd5e920d0ef641c684b/middleware/auth-utils/config.js#L131

keycloak.d.ts
bruegth: same as above

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'minTimeBetweenJwksRequests'?: number
```
bruegth, line 29:
```suggestion
    minTimeBetweenJwksRequests?: number
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
     'bearer-only'?: boolean
-    realm: string
+    'plainKey'?: string
+    'verifyTokenAudience'?: boolean
```
bruegth, line 32:
```suggestion
    verifyTokenAudience?: boolean
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
+    'clientId': string
+    'secret': string
+    'resource'?: string
+    'public-client'?: string
+    'realmUrl'?: string
+    'realmAdminUrl'?: string
+    'minTimeBetweenJwksRequests'?: number
     'bearer-only'?: boolean
-    realm: string
+    'plainKey'?: string
```
bruegth, line 31:
```suggestion
    realmPublicKey?: string
```

keycloak.d.ts
bruegth, line 26:
```suggestion
    public?: string
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
 declare namespace KeycloakConnect {
 
   interface KeycloakConfig {
-    'confidential-port': string|number
     'auth-server-url': string
```
bruegth, line 21:
```suggestion
    authServerUrl: string
```

keycloak.d.ts
```diff
@@ -18,12 +18,18 @@ interface KeycloakConnectStatic {
...
-    'ssl-required': string
+    'realm': string
+    'clientId': string
+    'secret': string
+    'resource'?: string
+    'public-client'?: string
+    'realmUrl'?: string
+    'realmAdminUrl'?: string
+    'minTimeBetweenJwksRequests'?: number
     'bearer-only'?: boolean
```
bruegth, line 30:
```suggestion
    bearerOnly?: boolean
```

bruegth (changes requested):
I think we should avoid using quotes around variables.

bruegth:
Typescript won't work with this package, maybe it's better to remove .d.ts completely until it's covered by unit tests.

c0bra:
What's the workaround for using this package with typescript in the meantime?

bruegth:
> What's the workaround for using this package with typescript in the meantime?

I've created a gulp task to replace the definition file in the node_modules folder with a modified one.

ItsTarik:
Hi and thanks for reporting,I've got an other similar issue! it's hard to find useful information about this issue since the typescript config `KeycloakConfig` is missing the realm-public-key prop already !
So that's what i get with both version `^11.0.3` and the actual one `^12.0.4`: 
```
  interface KeycloakConfig {
    'confidential-port': string|number
    'auth-server-url': string
    'resource': string
    'ssl-required': string
    'bearer-only'?: boolean
    realm: string
  }

```
So in my case i can use the missing declared property:` 'realm-public-key': 'pemkey'`, while ignoring the ts checking.
I will create a separate issue for this but i think that's important to list in this thread since its similar to the issue subject.

ffleandro:
What is the current status for Typescript support and what was the blocker on this PR?

PR #255, opened 2020-07-23
KEYCLOAK-13804 Require chromedriver only if it is not provided by env…

mhajas:
…ironment

This is rather a workaround than a fix, however, I would say for us should be enough. The failure was caused by failure to download/unzip chromedriver (see the second snippet in the Jira). The process of `npm install` continues because chromedriver is just optional dependency. Then when `webdriver.js` did `require('chromedriver')` it failed, because chromedriver was not correctly downloaded.

My changes execute `require('chromedriver')` only when it is really necessary, e.g. when chromedriver is not specified in the arguments/env variable. This change should be enough for us because, in CI, we use an environment variable to specifying chromedriver path.

### DIFF
--- a/test/utils/webdriver.js
+++ b/test/utils/webdriver.js
@@ -18,7 +18,6 @@
  * An utility with the specifics for selenium
  */
 const chrome = require('selenium-webdriver/chrome');
-const chromedriver = require('chromedriver');
 const webdriver = require('selenium-webdriver');
 const args = require('minimist')(process.argv.slice(2));
 const By = webdriver.By;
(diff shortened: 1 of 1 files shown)


mhajas:
CI results: https://keycloak-jenkins.com/job/universal-test-pipeline-adapters/1590/

mhajas:
@vmuzikar I think this is happening only with Node.js 10. 

To me, it seems like the issue is that npm downloads `chromedriver_linux.zip` into `/tmp/_version_/chromedriver/chromedriver_linux.zip` and then it tries to unzip it into `/tmp/_version_/chromedriver`. Then, there might be a file name conflict. The zip file contains a file with the name chromedriver, and also the directory is called chromedriver, so maybe during the unzip the tool confuses the destination `/tmp/_version_/chromedriver` and `/tmp/_version_/chromedriver/chromedriver`. If it chooses the first one, it fails because the directory is already there. However, I haven't found any info about this on the internet, so I am not sure it is correct.

PR #262, opened 2020-10-14
[KEYCLOAK-15950] [Ankon] Added option to set config on a keycloak ins…

luciferankon:


apiel:
This PR is inconsistent as the config are used in the instantiation of the Keycloak object, see https://github.com/keycloak/keycloak-nodejs-connect/blob/main/keycloak.js#L61

If you would want to do something like this, you should reinitialize the whole keycloak instance or extract all the config initialization and take care that the `GrantManager` is properly updated. I think to do such thing, you need to do major refactoring.

PR #263, opened 2020-10-20
KEYCLOAK-16026 Add tests for cookie store

vmuzikar:
I've reused the original JIRA ([KEYCLOAK-15859](https://issues.redhat.com/browse/KEYCLOAK-15859)) as it's not yet closed. Let me know if I should create a new JIRA just for this.

### DIFF
--- a/test/keycloak-connect-web-spec.js
+++ b/test/keycloak-connect-web-spec.js
@@ -285,6 +285,30 @@ test('Public client should work with slash in the end of auth-server-url', t => 
   });
 });
 
+test('App should be able to use cookie-store', t => {
+  t.plan(1);
+  var app = new NodeApp();
+  var client = admin.createClient(app.publicClient('appCookies'));
+
+  return client.then((installation) => {
+    app.build(installation, { cookies: true });
+    return page.get(app.port, '/cookie').then(() => {
+      page.login('alice', 'password').then(() => {
+        driver.navigate().refresh().then(() => {
+          page.events().getText().then(text => {
+            t.equal(text, 'Auth Success', 'User should be authenticated');
+          });
+        });
+      });
+    }).then(() => {
+      app.destroy();
+    }).catch(err => {
+      app.destroy();
+      throw err;
+    });
+  });
+});
+
 test('teardown', t => {
   return realmManager.then((realm) => {
     app.destroy();
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: focal
 language: node_js
 cache:
   npm: false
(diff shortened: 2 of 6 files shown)


vmuzikar:
@mhajas Could you please review one you get a chance?

slaskawi (dismissed):
LGTM. However, I'm definitely not an expert in JS. 

Maybe @ASzc and @stianst could review it?

vmuzikar:
I've fixed the linter errors.

@slaskawi Thanks for your review!

vmuzikar:
There seem to be some problem with Chrome in Travis. It apparently affects other PRs too. I'm trying to fix it.

vmuzikar:
Ok, Travis is now happy. We were using outdated distro which doesn't support latest Chrome.

Ready to (re-)review. :)

slaskawi (dismissed):

test/keycloak-connect-web-spec.js
```diff
@@ -285,6 +285,30 @@ test('Public client should work with slash in the end of auth-server-url', t =>
...
+  var app = new NodeApp();
+  var client = admin.createClient(app.publicClient('appCookies'));
+
+  return client.then((installation) => {
+    app.build(installation, { cookies: true });
+    return page.get(app.port, '/cookie').then(() => {
+      page.login('alice', 'password').then(() => {
+        driver.navigate().refresh().then(() => {
+          page.events().getText().then(text => {
+            t.equal(text, 'Auth Success', 'User should be authenticated');
```
mhajas, line 299: @vmuzikar should we check here that output element contains stuff from `keycloak-token` cookie? Or the presence of `Auth Success` is enough for us and implies that the cookies were readable?

vmuzikar: Good point. I believe the `Auth Success` should be enough. If the `keycloak-token` cookie wouldn't contain a valid input, it would fail on JSON parsing here:
https://github.com/keycloak/keycloak-nodejs-connect/pull/263/files#diff-2c6e47101e422ddc2041c33fe87447ffabb5da93c296e0d89426b91cef5d75feR132

mhajas: I quickly tried this: 
```
console.log(JSON.parse(null))
```
And it works, but I am not sure what `req.cookies['keycloak-token']` returns when there is no `keycloak-token` cookie.

vmuzikar: It should return `undefined` if the cookie is missing (I accidentally experienced it myself when preparing the fix).

But to be on the safe side, I've added a small if:
https://github.com/keycloak/keycloak-nodejs-connect/blob/fd245b43359b7790319544ab230b221495a6e4e9/test/fixtures/node-console/index.js#L132

mhajas (approved):
Thanks for the explanation and fix @vmuzikar, LGTM.

abstractj (approved):

abstractj:
Thanks @vmuzikar for taking care of this and thanks @mhajas @slaskawi for reviewing.

PR #270, opened 2020-12-30
[KEYCLOAK-16726] fix redirect for restify

kapekost:
### Using restify in node-connect, fails to redirect the client with the following error

`Exception has occurred: AssertionError [ERR_ASSERTION]: res.redirect() requires a next param (func) is required`
Alongside with this, I noticed some params from the request object are not collected correctly, that'd be the host, protocol and port. 

And took the chance to update the deprecated method for express.js: redirect now needs a `status code`

### To reproduce: 
 - use restify instead of express js
 - call the protect() function in a request

#### JIRA Ticket:
 https://issues.redhat.com/browse/KEYCLOAK-16726

### DIFF
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -33,10 +33,10 @@ function forceLogin (keycloak, request, response, next) {
   let uuid = UUID();
   let loginURL = keycloak.loginUrl(uuid, redirectUrl);
   try {
-    // expressjs not happy with next param
+    // express.js not happy with next param
     response.redirect(301, loginURL);
   } catch (ex) {
-    console.log('using Restify? expected error. Switching redirect call ..');
+    console.debug('using Restify? expected error. Switching redirect call ..', ex);
     // for restify mandatory next param
     response.redirect(301, loginURL, next);
   }
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -17,11 +17,11 @@
 
 const UUID = require('./../uuid');
 
-function forceLogin (keycloak, request, response) {
-  let host = request.hostname;
+function forceLogin (keycloak, request, response, next) {
   let headerHost = request.headers.host.split(':');
+  let host = request.hostname || headerHost[0];
   let port = headerHost[1] || '';
-  let protocol = request.protocol;
+  let protocol = (request.protocol !== undefined) ? request.protocol : request.isSecure() ? 'https' : 'http';
   let hasQuery = ~(request.originalUrl || request.url).indexOf('?');
 
   let redirectUrl = protocol + '://' + host + (port === '' ? '' : ':' + port) + (request.originalUrl || request.url) + (hasQuery ? '&' : '?') + 'auth_callback=1';
@@ -32,7 +32,14 @@ function forceLogin (keycloak, request, response) {
 
   let uuid = UUID();
   let loginURL = keycloak.loginUrl(uuid, redirectUrl);
-  response.redirect(loginURL);
+  try {
+    // expressjs not happy with next param
+    response.redirect(301, loginURL);
+  } catch (ex) {
+    console.log('using Restify? expected error. Switching redirect call ..');
+    // for restify mandatory next param
+    response.redirect(301, loginURL, next);
+  }
 }
 
 function simpleGuard (role, token) {
@@ -58,7 +65,7 @@ module.exports = function (keycloak, spec) {
     }
 
     if (keycloak.redirectToLogin(request)) {
-      forceLogin(keycloak, request, response);
+      forceLogin(keycloak, request, response, next);
     } else {
       return keycloak.accessDenied(request, response, next);
     }
--- a/middleware/protect.js
+++ b/middleware/protect.js
@@ -34,11 +34,11 @@ function forceLogin (keycloak, request, response, next) {
   let loginURL = keycloak.loginUrl(uuid, redirectUrl);
   try {
     // express.js not happy with next param
-    response.redirect(301, loginURL);
+    response.redirect(302, loginURL);
   } catch (ex) {
-    console.debug('using Restify? expected error. Switching redirect call ..', ex);
+    console.debug('Using Restify? Expected error. Switching redirect call ...', ex);
     // for restify mandatory next param
-    response.redirect(301, loginURL, next);
+    response.redirect(302, loginURL, next);
   }
 }
 
--- a/test/keycloak-connect-rest-mixed-client-restify-spec.js
+++ b/test/keycloak-connect-rest-mixed-client-restify-spec.js
@@ -1,0 +1,107 @@
+'use strict';
+
+const admin = require('./utils/realm');
+const NodeApp = require('./fixtures/node-console/restify').NodeApp;
+
+const test = require('blue-tape');
+const axios = require('axios');
+const getToken = require('./utils/token');
+
+const realmName = 'mixed-mode-realm';
+const realmManager = admin.createRealm(realmName);
+const app = new NodeApp();
+
+const auth = {
+  username: 'test-admin',
+  password: 'password'
+};
+
+test('setup', t => {
+  return realmManager.then(() => {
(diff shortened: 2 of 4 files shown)


abstractj:
@kapekost please add the integration tests for it.

kapekost:
hi @abstractj  added tests :) I verified that if I remove the fix it fails with the error regarding 'next'

btw, I couldn't manage to run the tests isolated, I am getting an odd error that it cannot find module resolve ... I tried to install it globally and locally ..nothing.. might be a node version or other version miss-match?

kapekost:
@abstractj any thoughts on this?

ekoeryanto:
hi, @kapekost. is #275 fix this?

kapekost:
hey, thanks for asking, looks nice
but
I just cherry picked your code and ran my tests:
```
-------------------------------

TAP version 13
# setup
# Should test protected route.
/home/kapekost/dev/kapekost/keycloak-nodejs-connect/middleware/protect.js:36
  response.writeHead(302, { Location: loginURL }).end();
                                                 ^

TypeError: Cannot read property 'end' of undefined
    at forceLogin (/home/kapekost/dev/kapekost/keycloak-nodejs-connect/middleware/protect.js:36:50)
    at protect (/home/kapekost/dev/kapekost/keycloak-nodejs-connect/middleware/protect.js:62:7)
    at nextTick (/home/kapekost/dev/kapekost/keycloak-nodejs-connect/node_modules/restify/lib/chain.js:167:13)
    at processTicksAndRejections (internal/process/task_queues.js:75:11)
```

looks like `restify` is not happy with this format, also the http/https identification needs adjustment like the one I have in my branch ( I couldn't find a more elegant way for it)

maybe you can use my branch to add a combination of your solution to mine, that would also run my extended tests

ekoeryanto:
@kapekost, can you use it without chaining the .end method?
```js
response.writeHead(...)
response.end()
```

kapekost:
I can give it a go tomorrow, though it will still need the patches to call next() instead

kapekost:
Looks like sendRaw might work

_sendRaw
Like res.send(), but skips formatting. This can be useful when the payload has already been preformatted. Sends the response object. pass through to internal __send that skips formatters entirely and sends the content as is._

middleware/protect.js
```diff
@@ -17,11 +17,11 @@
...
 const UUID = require('./../uuid');
 
-function forceLogin (keycloak, request, response) {
-  let host = request.hostname;
+function forceLogin (keycloak, request, response, next) {
   let headerHost = request.headers.host.split(':');
+  let host = request.hostname || headerHost[0];
   let port = headerHost[1] || '';
-  let protocol = request.protocol;
+  let protocol = (request.protocol !== undefined) ? request.protocol : request.isSecure() ? 'https' : 'http';
```
ekoeryanto, line 24: does `request.isSecure` exist on node http or express?
what about `request.connection.secure`?
what if the application is behind proxy?

ekoeryanto (changes requested):

middleware/protect.js
kapekost: that was my main headache on the patch.. seems like `protocol` comes on `express`, therefore I left it to apply as it was before the patch. 

`isSecure()` is wrapping the `encrypted`, based on this [doc](http://restify.com/docs/request-api/#issecure), 
should cover the cases behind proxy for `restify` too, 

more or less I used that toggle in that line to somehow tell us if it's `restify` or not, I couldn't find a prettier way to do that..

abstractj:
@kapekost thanks for submitting this. Do we have any thread on the dev mailing list about including restify support for the Node.js adapter?

kapekost:
You're welcome :) 
I'm not aware about a mail list sorry, 
just submitted the patch I'm using that seems to make it work so far, 
I saw calls to next() in there, just missing this one in redirect(), maybe due to lib updates whatsoever.. I'd guess you used to support it and somehow it got left behind?

mhajas (commented):
@kapekost Thank you very much for your PR, the changes look reasonable to me. All tests are passing for me locally. However, I am not sure I am skilled enough to think through all aspects of this PR. @abstractj Should we ask for some reviews on community ML?

middleware/protect.js
```diff
@@ -32,7 +32,14 @@ function forceLogin (keycloak, request, response) {
 
   let uuid = UUID();
   let loginURL = keycloak.loginUrl(uuid, redirectUrl);
-  response.redirect(loginURL);
+  try {
+    // express.js not happy with next param
+    response.redirect(301, loginURL);
```
vmuzikar, line 37: I think we should use `302`. Looking into the [docs](http://expressjs.com/en/5x/api.html#res.redirect) and I think this is what we used to return before as it's the default value. Also IMHO `302` (found a.k.a moved temporarily) is more appropriate than `301` (moved permanently).

middleware/protect.js
```diff
@@ -32,7 +32,14 @@ function forceLogin (keycloak, request, response) {
...
+  } catch (ex) {
+    console.debug('using Restify? expected error. Switching redirect call ..', ex);
+    // for restify mandatory next param
+    response.redirect(301, loginURL, next);
```
vmuzikar, line 41: Same as above.

test/keycloak-connect-rest-mixed-client-restify-spec.js
```diff
@@ -0,0 +1,107 @@
...
+const realmName = 'mixed-mode-realm';
+const realmManager = admin.createRealm(realmName);
+const app = new NodeApp();
+
+const auth = {
+  username: 'test-admin',
+  password: 'password'
+};
+
+test('setup', t => {
```
vmuzikar, line 19: Tests look good but I noticed most of the code (both test app and tests) are copy-pasted. Couldn't we somehow make it more generic? More duplicate code means higher maintenance costs.

vmuzikar (changes requested):
@kapekost Thank you for your contribution! Added two comments.

middleware/protect.js
kapekost: sure, it does make sense, I will push an update for both of them

test/keycloak-connect-rest-mixed-client-restify-spec.js
kapekost: indeed, it was tough to get my head around, to the minimum required level to strip out the basic tests + setup and swap express with restify. I can dig deeper to think of a way to make a toggle, but I am afraid I am not familiar enough with this setup, and it will be hard to find enough time, I guess another option would be to strip it down even further? And just make sure this copy of tests will only hit that redirect at least once?

middleware/protect.js
kapekost: Having second thoughts on this, If it's permanent redirect , it might skip browser checking the destination url that we modify on a next call 🤔 what do you think?

vmuzikar: Well of course, that's why we want to use `302`. 😉

test/keycloak-connect-rest-mixed-client-restify-spec.js
vmuzikar: I don't mind stripping out the restify test (and app) even more. But I'm wondering whether it wouldn't be worth it to run all the tests with restify (together with express) to prevent issues like this in the future. @abstractj Are you sure we don't support restify? Looking into the [docs](https://www.keycloak.org/docs/latest/securing_apps/index.html#_nodejs_adapter) and it says we support frameworks _like_ express.

kapekost: indeed, currently it looks like creating blockers, there was a nice suggestion (can't remember by whom) to pass in  the redirect command (at lest url) Constructed on the outer side... I just got a case failing to identify correctly http vs https ...

kapekost:
Looks like there are more cases using redirect in a different way, therefore this is not a complete "restify support" 

so I am closing this PR, if the development team decides to fully support other frameworks, they have a hint of the issue :)

PR #282, opened 2021-05-16
[KEYCLOAK-18049] - Update for Windows to run tests and linting

acotty:
Kycloak issue: https://issues.redhat.com/browse/KEYCLOAK-18049

The changes are to enable Windows developers to run the unit tests and linting on a standard Windows node dev type environment 
going forward as the current way of testing and linting does not work on a Windows without installing third party utilities (e.g. Cygwin), 
which is not required for any other NodeJS package I have not come across so far with the exception of Docker/Docker Compose.

a) Reworking the dev package dependencies. Updated packages and removed redundant packages.
b) Effectively migrating scripts into the package.json and update readme.md. Removed requirement for bash shell for most developers (still required for Redhat release process for some resson).
c) Rework the way docker is started in order to use a docker-compose.yml file instead of a script. Now will work on all platforms supporting docker. Also removed make.exe requirement.
d) Rework unit tests that fail due to running on latest versions of nodejs on Windows due to various issues. Tests run under Windows and Ubuntu locally and in ubuntu on github actions.
e) Update the travis CI and github actions to use docker-compose.yml file and "npm run lint" and "npm run test" and remove NodJS 11 & 13 from matrix, added NodJS 14 & 16 to matrix. 
g) The tests now use NYC to produce and check the code coverage for a pass or fail.
h) The dev setup and test process now works on Windows after the changes as it did not before the changes.

The changes do NOT fix the following:
1) Ensure all of the tests are written consistently.
2) Page loading issues with Sellenium where code changes need to be done to ensure delays are dynamic instead of fixed.
3) Timeouts due to testing on slow environments (aka Travis or Github compared to local)
4) Documentation on testing and code changes are 100% perfect or comprehensive enough for every developer.

To see the changes running in github actions please have a look at 

https://github.com/acotty/keycloak-nodejs-connect/runs/2592862415?check_suite_focus=true

This is a new PR as I could not open the previous one and have updated the request info above so it gives more details of the changes and reasons. Failure and pass details are included in the redhat issue.


docker-compose-keycloak-mysql.yml
```diff
@@ -0,0 +1,160 @@
+version: "3.7"
+
+x-logging:
```
fidgi, line 3: IHMO it is better to keep this outside the docker-compose file , to let people adjust log seeting according to their env

docker-compose-keycloak-mysql.yml
```diff
@@ -0,0 +1,160 @@
...
+      MYSQL_DATABASE: keycloak
+    healthcheck:
+      test: "mysqladmin ping -u root -p$${MYSQL_ROOT_PASSWORD}"
+      start_period: 10s
+      timeout: 20s
+      retries: 10
+    volumes:
+      # On Windows if your Keycloak times out setting up the DB then try to put the volume on a faster drive (C: is an SSD!!!)
+      #- ./DockerVolumes/mysql-data:/var/lib/mysql
+      - C:/docker/volumes/mysql-keycloak-testing:/var/lib/mysql
```
fidgi, line 28: Too much Windows oriented

docker-compose-keycloak-mysql.yml
```diff
@@ -0,0 +1,160 @@
...
+    restart: unless-stopped
+    depends_on:
+      - mysql
+    ports:
+      - 8081:8080    
+    networks:
+      - docker-network
+    logging: *default-logging      
+    
+  portainer:
```
fidgi, line 50: Not everyone wants portainer to be included

docker-compose.yml
```diff
@@ -0,0 +1,107 @@
...
+        #- \\.\pipe\docker_engine:\\.\pipe\docker_engine
+        #-C:\ProgramData\Portainer:C:\data 
+    ports:
+      - 8000:8000
+      - 9000:9000
+    networks:
+      - docker-network
+    logging: *default-logging      
+
+  keycloak_SA:
```
fidgi, line 36: why not just "keycloak" ?

keycloak.js
```diff
@@ -60,6 +60,10 @@ var CheckSso = require('./middleware/check-sso');
  */
 function Keycloak (config, keycloakConfig) {
   // If keycloakConfig is null, Config() will search for `keycloak.json`.
+  if (keycloakConfig === undefined)
+  {
+    console.error(`Keycloak called with no keycloakConfig specified`);
```
fidgi, line 65: Please, do not add console output

package.json
```diff
@@ -35,46 +38,43 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "jwk-to-pem": "^2.0.0"
+    "jwk-to-pem": "^2.0.5",
+    "keycloak-admin-client": "bucharest-gold/keycloak-admin-client"
```
fidgi, line 42: Why do you need to move this dependency from dev ?

test/grant-manager-token-timeout-spec.js
```diff
@@ -2,54 +2,86 @@
 
 const GrantManager = require('../middleware/auth-utils/grant-manager');
 const Config = require('../middleware/auth-utils/config');
-const test = require('tape');
+const t = require('tap');
```
fidgi, line 5: IHMO, test is better

docker-compose-keycloak-mysql.yml
acotty: I have been thinking about the docker-compose-keycloak-mysql.yml file and it is probably better for me to remove this file and submit a PR later to add it as an example with code and do the following:
1) Rename the existing example directory to examples
2) Move the existing example files into an examples/web_app_simple directory
3) Create another example directory like examples/web_app_complex that uses the docker-compose-keycloak-mysql.yml  to show how to setup and use keycloak with a MYSQL server

docker-compose-keycloak-mysql.yml
acotty: See first comment response.

docker-compose-keycloak-mysql.yml
acotty: See first comment response.

docker-compose.yml
acotty: As it is standalone and does not required any other containers to run.

keycloak.js
acotty: I was following the exiting code, which includes a bunch of console.log and console.error calls. If you want this removed I can, but I added it as I spent about 2 hours tracking down a call which ended up calling this function without a keycloakConfig and thought it may be a good idea to add this.

The better way is to rework ALL of the console calls in the code (apart from maybe the tests) to use the NPMJS debug package, which would allow devs to conditionally turn on/off debug output. But this is definitly would not be part of this PR.

Let me know if you want me to leave this in or remove it.

package.json
acotty: I can't remember, but it may be that I spotted it in the product/prod-arguments.json and as such thought that this was used in the Redhat release process (I cannot find any docs or have seen any on what the release scripts and files are used for/how BTW).

test/grant-manager-token-timeout-spec.js
acotty: If you want this then have a look at the attached file as it shows what the code would look like and it's not nice IMHO as it would require all of the references in the code to 't' to become 'test' and as such would mean allot more changes and make the code potentially harder
[test.js.txt](https://github.com/keycloak/keycloak-nodejs-connect/files/6493133/test.js.txt)
 to understand as 't' and 'test' both become 'test'

acotty:
Let me know if the comments are okay and if they are I can make the following two changes and push the code up:
1) Change to the package.json  - move keycloak-admin-client back to dev.
2) Remove the docker-compose-keycloak-mysql.yml

Thanks,
Andrew

abstractj:
@acotty @fidgi the number of changes in this PR makes it really hard to review. Few considerations:

1. If you would like to make it supported on Windows, please, do not introduce additional changes like: refactoring constants, upgrading dependencies, docker composer files, or adding log messages that we don't need...
2. It must be tested on GH actions first. So, a good first step would be to include Windows as a platform on GH actions. For example:
```
strategy:
      matrix:
        platform: [ubuntu-latest, macos-latest, windows-latest]
```

For now, I really appreciate you taking the time to make this patch. However, these changes were not discussed before writing it, but I think it would be more effective if we take one step back. 

For now, I'm closing this issue but would love to see you open a patch that takes the above direction:
- Add Windows as part of GH actions matrix
- If tests failed, submit a PR with the fix if you would like.

acotty:
@abstractj I will leave it for you to add in the os matrix to githuib actions as currently docker in the Github virtual environments do not support linux containers and keycloak does not have a windows container as you know.

As for closing the PR. if you want to discuss it I suggest that instead of closing PR's or issues you add comments so it can be discussed instead of just closing the PR or issue before any discussion as closing it after a single comment is NOT a discussion.

acotty:
Pull requests were submitted, BUT were put on the future plan on the Redhat site, so they have been closed as they would all be stale by the time abstratj got around to looking at them.....

PR #284, opened 2021-05-19
KEYCLOAK-18130 - Migrate to eslint from jshint so linting can work on non Unix OS

acotty:
https://issues.redhat.com/browse/KEYCLOAK-18130

**KEYCLOAK-18130 Description - relevant info**
_The node package jshint does not glob on Windows and therefore does not work on Windows, which causes the "npm run lint" to fail on windows. There is no workar round for this without changing to another linter or  making changes to jshint in a PR.

There are a few of issues already raised about windows globing on https://github.com/jshint/jshint/issues   , but these have not been resolved w.r.t. fixing them.

 

I suggest moving to eslint as it works on windows, Linux and macOS. If this is okay please let me know and I will do the work required just to to fix running "npm run lint" on a windows on the command line._

So this PR is to migrate from jshint to eslint.  Please be aware that the .eslintrc.json file may contain linting conditions that IMHO should be enabled, but are disabled/downgraded to warnings in order to match the faults/non faults semistandard fails on.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,36 @@
+{
+  "env": {
+    "node": true,
+    "browser": true,
+    "es6": true
+  },
+  "parserOptions": {
+     "ecmaFeatures": {
+       "jsx": true
+    },
+    "ecmaVersion": 8,
+    "sourceType": "module"
+  },
+  "globals": {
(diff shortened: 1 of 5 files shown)


acotty:

PR #286, opened 2021-05-20
KEYCLOAK-18133 - fix "npm run docs"

acotty:
https://issues.redhat.com/browse/KEYCLOAK-18133
Package.json "docs" script fails due to non existent files in the jsdoc parameter list and glob issues.

I propose to fix this by:

1) adding a .jsdoc.json file

2) Changing the package.json docs script to be:

"docsNew": "rimraf docs-js && jsdoc -c .jsdoc.json",

3) Add rimraf for it's ability to delete directory trees on multiple platforms and file systems.

ALSO when implementing this the docs now go into the docs-js directory instead of into the existing docs directory. This indicates to me that no one has looked at the docs creation for a long time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ pid.txt
 .idea/
 tmp/
 .vscode/
+docs-js/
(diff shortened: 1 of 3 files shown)


acotty:

PR #287, opened 2021-05-20
KEYCLOAK-18131 Add support for Windows to spin up docker instead of using bash as there is no bash installed by default

acotty:
https://issues.redhat.com/browse/KEYCLOAK-18131

By default Windows does not include a bash shell and as such you cannot run by default either of the following scripts on Windows:

scripts/start-server.sh
scripts/stop-server.sh

 
I propose to add a docker-compose/yml and update the docs to add in the appropriate commands to create the docker keycloak container and also stop & destroy it


acotty:

PR #288, opened 2021-05-20
KEYCLOAK-18132 - Enable tests to be run on on a standard Windows OS install

acotty:
https://issues.redhat.com/browse/KEYCLOAK-18132
By default Windows does not include a bash shell and as such you cannot run by default the run-tests.sh on Windows that is used to run the tests.

 

I propose to do the following:

1) add another script enttry in the package.json to directly run the scripts on Windows so that the tests can be fixed in a future issue. The line to be added would potentially be:

"testonWindows": "./node_modules/.bin/tape --timeout=360 -j1 test/*-spec.js",

2) Update the tests-development.md to change the following line

make tests

to

make tests

or on windows run "npm run testonWindows"


acotty:

PR #290, opened 2021-05-24
updates and improvements to TS typings

evanshortiss:
This PR add types to safely access to the underlying `kauth` object and read token data.

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -1,10 +1,25 @@
 import * as express from 'express'
 
+declare global {
+  namespace Express {
+    // Add the kauth property to the incoming express.Request object
+    interface Request {
+      /**
+       * Supports accessing token and session information, e.g:
+       * req.kauth.grant.access_token?.content.name returns the user's name
+       */
+      kauth: {
+        grant: KeycloakConnect.Grant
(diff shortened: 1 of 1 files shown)


keycloak.d.ts
```diff
@@ -1,10 +1,25 @@
...
+declare global {
+  namespace Express {
+    // Add the kauth property to the incoming express.Request object
+    interface Request {
+      /**
+       * Supports accessing token and session information, e.g:
+       * req.kauth.grant.access_token?.content.name returns the user's name
+       */
+      kauth: {
+        grant: KeycloakConnect.Grant
```
evanshortiss, line 12: The grant refers to properties such as `grant.access_token`. Will these *always* be defined? Current typings show them as possibly undefined (using the `?` operator) as seen [here](https://github.com/keycloak/keycloak-nodejs-connect/blob/master/keycloak.d.ts#L35-L41).

Is it safe to remove the `?` operator from those?

edewit:
No need to open a JIRA we are using github issues now

helio-frota:
> No need to open a JIRA we are using github issues now

@edewit that is a great move :+1:

PR #294, opened 2021-07-29
[KEYCLOAK-17490] Add override ISS

ATeal:
This feature will allow you to override the ISS validation.

Currently the library isn't respecting the `FRONTEND_URL` configurationof keycloak which will give you a different issuer than the backend server url

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -20,6 +20,7 @@ declare namespace KeycloakConnect {
   interface KeycloakConfig {
     'confidential-port': string|number
     'auth-server-url': string
+    'iss' : string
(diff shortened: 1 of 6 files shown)


RaqieT:
I think ISS should be picked up from body (field: "issuer") of `/auth/realms/{realm}/.well-known/openid-configuration` endpoint on adapter startup, so you don't have to provide ISS in your backend app. I've already seen this approach in diffrent keycloak adapters (for example: Spring Boot adapter).

elirenato:
+1

tskorupka:
Would there be a possibility to merge this?

PR #297, opened 2021-09-20
 KEYCLOAK-19343 propagate error message from token validation to calling code

velias:
`keycloak.getGrant()` method does not propagate error message from `grantManager.createGrant()` but it creates new error with generic message "Could not store grant code error", which is misleading and doesn't help to troubleshoot what is going on in case of problems. `createGrant()` perform token validation and there may be many reasons why it can fail and it is necessary to see what is going.

### DIFF
--- a/keycloak.js
+++ b/keycloak.js
@@ -326,8 +326,7 @@ Keycloak.prototype.getGrant = function (request, response) {
       .then(grant => {
         self.storeGrant(grant, request, response);
         return grant;
-      })
-      .catch(() => { return Promise.reject(new Error('Could not store grant code error')); });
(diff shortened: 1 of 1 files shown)


velias:
Hi @abstractj, any chance to get this merged? Very small change but big improvement for real life usage and for support staff. Thanks in advance.

abstractj:
@velias apologies for the long delay, but to merge it, we need tests attached to those changes.

abstractj:
Thanks for submitting those changes, unfortunately due to the lack of activity I'm closing it. But if you decide to update those changes in the future, please reopen.

PR #313, opened 2021-12-21
Update with the reference to the version of the keycloak-admin-client without breaking backward compatibility

### LINKED ISSUE #307: Incorrect dependency in package.json
### Describe the bug

In the package.json file following dependency is defined with Github source

"keycloak-admin-client": "bucharest-gold/keycloak-admin-client"

but this repository is either not present or not public. This is breaking the example.

### Version

16.0.0-dev

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

### Anything else?

_No response_

abstractj:
Closes #307

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "ink-docstrap": "^1.1.4",
     "jsdoc": "^3.6.3",
     "jshint": "^2.10.2",
-    "keycloak-admin-client": "bucharest-gold/keycloak-admin-client",
+    "keycloak-admin-client": "^0.12.0",
     "keycloak-request-token": "^0.1.0",
     "nock": "^12.0.3",
     "nyc": "^15.0.0",


package.json
```diff
@@ -49,7 +49,7 @@
     "ink-docstrap": "^1.1.4",
     "jsdoc": "^3.6.3",
     "jshint": "^2.10.2",
-    "keycloak-admin-client": "bucharest-gold/keycloak-admin-client",
+    "keycloak-admin-client": "abstractj/keycloak-admin-client",
```
jonkoops, line 52: Why not use the latest version of [`keycloak-admin-client`](https://www.npmjs.com/package/keycloak-admin-client) that was published on NPM? Does this repo contain changes that are missing there?

jonkoops (changes requested):

package.json
abstractj: @jonkoops because I missed that on npm :) Thanks for that, PR updated. If all the checks pass, I will merge it so we can discuss more about the migration next year, and I don't have to bother you again.

abstractj:
Quay is experiencing some issues https://status.quay.io/ and that's the reason why the build is failing. Once they fix it, we can trigger the job again.

jonkoops (commented):
Retroactive approval, but things are looking good here :+1:

edewit:
the keycloak admin client alternative: #315

PR #315, opened 2021-12-23
Changed to use keycloak-admin-client

edewit:
Now uses the `@keycloak/keycloak-admin-client` instead


test/utils/realm.js
```diff
@@ -33,8 +34,10 @@ var kca = keycloakAdminClient(settings);
  */
 function createRealm (realmName) {
   var name = realmName || 'test-realm';
-  return kca.then((client) => {
-    return client.realms.create(parse(realmTemplate, name));
+  return kca.auth(settings).then(() => {
+    return kca.realms.create(parse(realmTemplate, name)).then(() => {
+      return kca.realms.findOne({ realm: realmName });
+    });
```
jonkoops, line 40: I believe that sometimes the realm name can be sanitized by the API when created, so perhaps we can use the `realmName` that is returned from the endpoint?
```suggestion
    return kca.realms.create(parse(realmTemplate, name)).then(({ realmName: realm }) => {
      return kca.realms.findOne({ realm });
    });
```

That said, does the `create()` call not return the entire `RealmRepresentation`? If so we can simply omit the `findOne()` call.

package.json
```diff
@@ -38,18 +38,18 @@
     "jwk-to-pem": "^2.0.0"
```
jonkoops, line 38: Can we also check in the package lock file to keep the builds reproducible?

jonkoops (approved):
Couple of small comments, but nothing that I would consider blocking.

abstractj (approved):
LGTM thank you, @edewit 

Waiting for @jonkoops approval before merging it.

test/utils/realm.js
jonkoops: Since we're only using this for tests let's keep things as they are.

jonkoops (approved):

PR #318, opened 2022-01-07
Remove .npmignore in favor of 'files' field

jonkoops:
We're already using the 'files' field in the package to explicitly include the files in the final package, so this file can be removed.


abstractj (approved):
LGTM

edewit (approved):

PR #322, opened 2022-01-07
Drop Maven build from project

jonkoops:
Drops the Maven build from the project in order to reduce the amount of code we need to maintain. This build was needed for downstream RH-SSO, but it should now be possible to remove this as support for NPM packages was introduced.

@ASzc will have to confirm that we can indeed remove this code before merging this PR.


ASzc:
I'm investigating. It's possible we may need to keep `prod-arguments.json` (moved to repo root).

drichtarik:
I think the whole product dir can be deleted.

ASzc:
As far as I can tell, prod-arguments is still read by the automation for this build, and used to construct the PNC build config

drichtarik:
I think the pipeline sets up the PNC build configuration from the prod arguments, but we don't use that pipeline anymore and the prod arguments are obsolete. If you go to the latest pnc build of nodejs project, you can see that that data is different now because its not a wrapper anymore. Apart from that, I don't know where else it is used.

ASzc:
I was seeing the pipeline still called `pnc-update-build-cfg.sh`, which accesses prod-arugments, but if we don't use that pipeline anymore, then we can delete prod-arguments along with the rest

jonkoops:
@ASzc @drichtarik So what do y'all think can we merge this PR as is?

ASzc (approved):
Yeah, ok to merge

abstractj (approved):
LGTM

edewit (approved):

PR #323, opened 2022-01-10
provide options to cookie storage

apiel:
Allow user to provide custom options for cookie storage like httpOnly mode or custom expiration date:

```ts
const keycloak = new Keycloak(
        { 
            cookies: true,
            cookieOptions: { maxAge: 900000, httpOnly: true },
        },
        {
            realm: 'example',
            // ...
        }
};
```

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -30,6 +30,7 @@ declare namespace KeycloakConnect {
     scope?: string
     store?: any
     cookies?: boolean
+    cookieOptions?: express.CookieOptions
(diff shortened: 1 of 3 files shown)


oleaasbo:
I would like to see this added as well. A workaround for me has been:
```
		app.use(function (req, res, next) {
			var org = res.cookie;
			res.cookie = function (name, value, options) {
				// Change options here
				org.call(this, name, value, options);
			};
			next()
		})
```

PR #325, opened 2022-01-20
chore: use same LTS node versions as declared on CI

helio-frota:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "tape-catch": "^1.0.6"
       },
       "engines": {
-        "node": ">=4.6.2"
+        "node": ">=12"
       },
       "optionalDependencies": {
         "chromedriver": "latest"
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     ]
   },
   "engines": {
-    "node": ">=4.6.2"
+    "node": ">=12"
   },
   "repository": {
     "type": "git",


package.json
```diff
@@ -70,7 +70,7 @@
     ]
   },
   "engines": {
-    "node": ">=4.6.2"
+    "node": "^16 || ^14 || ^12"
```
jonkoops, line 73: This would cause the installation to break if folks are using Yarn on other versions of Node (such as Node 17). I'd recommend that we add our lowest target here instead.
```suggestion
    "node": ">=16"
```

jonkoops (changes requested):

package.json
helio-frota: makes sense to me. PR updated to `>=12` thank you for the review

package-lock.json
```diff
@@ -36,7 +36,7 @@
         "tape-catch": "^1.0.6"
       },
       "engines": {
-        "node": ">=4.6.2"
+        "node": "^16 || ^14 || ^12"
```
jonkoops, line 39: Can you also update the lockfile? An `npm install` should do it for you.

jonkoops (changes requested):

package-lock.json
helio-frota: sure thing, sorry I forgot:+1:  PR updated

jonkoops (approved):

edewit (approved):

PR #97, opened 2017-10-31
KEYCLOAK-5771

beau6183:
Introduces the option to return a promise from `keycloak.protect` in order to facilitate async operations in the middleware.

PR #355, opened 2022-03-07
add support for async guard function

jseparovic:
Reviving the following old PR as I'm hitting the same issue:
https://github.com/keycloak/keycloak-nodejs-connect/pull/97
https://issues.redhat.com/browse/KEYCLOAK-5771?workflowName=GIT+Pull+Request+workflow+&stepId=4

I need to use an async GuardFn for my application.

I tested with a simple setTimeout method:
```
const keycloakGuardFn: KeycloakConnect.GuardFn = async (accessToken: Token, req: express.Request): Promise<boolean> => {
    const sleep = async (ms: number) => {
        return new Promise((resolve) => setTimeout(resolve, ms));
    }
    logger.trace("Sleeping for 2 seconds");
    await sleep(2000);
    logger.trace("Sleep finished");
    return true;
}
```

Looks fine in my testing. 

Do I need to update the unit tests? I'm not too sure where/what to add here.

Cheers

### DIFF
--- a/keycloak.d.ts
+++ b/keycloak.d.ts
@@ -192,7 +192,7 @@ declare namespace KeycloakConnect {
     isExpired(): boolean
   }
 
-  type GuardFn = (accessToken: Token, req: express.Request, res: express.Response) => boolean
+  type GuardFn = (accessToken: Token, req: express.Request, res: express.Response) => boolean|Promise<boolean>
 
   interface EnforcerOptions {
     response_mode?: string,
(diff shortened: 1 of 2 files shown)


jseparovic:
Any chance of merging this in? What needs to be done on my end? Cheers

pjamroziak:
@jseparovic 
Library will be deprecated :( I don't think, if they will merge your changes :/

PR #359, opened 2022-03-25
Adjust nodejs adapter with OIDC RP-Initiated Logout

### LINKED ISSUE #358: Adjust nodejs adapter with OIDC RP-Initiated Logout
### Description

The Keycloak issue https://github.com/keycloak/keycloak/issues/10885 adds support for OIDC RP-Initiated logout to Keycloak. There are some changes in the logout logic, which caused some tests to fail in keycloak-nodejs-connect.

This task is to adjust with latest Keycloak main once the OIDC RP-Initiated logout is updated in Keycloak. The PR for keycloak-nodejs-connect can be merged after this PR to keycloak-main is merged: https://github.com/keycloak/keycloak/pull/10887

### Discussion

_No response_

### Motivation

_No response_

### Details

_No response_

mposolda:
Closes #358


keycloak.js
```diff
@@ -400,10 +400,17 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
   return url;
 };
 
-Keycloak.prototype.logoutUrl = function (redirectUrl) {
-  return this.config.realmUrl +
-  '/protocol/openid-connect/logout' +
-  '?redirect_uri=' + encodeURIComponent(redirectUrl);
+Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
```
jonkoops, line 403: You can use the [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) object to construct the query string. This deals with encoding keys, values and separators for you.

jonkoops, line 403: Can you also add this new parameter to the TypeScript definitions here? https://github.com/keycloak/keycloak-nodejs-connect/blob/773e95a2fd8b4046cb22f6e3d55853169a81e9a8/keycloak.d.ts#L416

jonkoops (changes requested):

mposolda:
@jonkoops Thanks for the review! I've addressed your comments

keycloak.js
```diff
@@ -400,10 +400,19 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
...
+  if (idTokenHint) {
+    let urlParams = new URLSearchParams();
+    urlParams.append("id_token_hint", idTokenHint);
+    if (redirectUrl) {
+      urlParams.append("post_logout_redirect_uri", redirectUrl);
+    }
+    url += "?" + urlParams.toString();
+  }
+  return url;
 };
```
jonkoops, line 416: Alternate proposal, we could construct a [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) and add the search params there. The only requirement would be that `config.realmUrl` has a valid structure, which means it needs at least a protocol and a host name.

```js
Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
  const url = new URL(this.config.realmUrl + '/protocol/openid-connect/logout');

  url.searchParams.set('post_logout_redirect_uri', redirectUrl);

  if (idTokenHint) {
    url.searchParams.set('id_token_hint', idTokenHint);
  }

  return url.toString();
};
```

package-lock.json
```diff
@@ -1,7313 +1,8 @@
 {
```
jonkoops, line 1: Since the dependencies have not changed this file should not be updated. It was updated locally because you are a running an older version of NPM and Node.

jonkoops (changes requested):

package-lock.json
mposolda: Good catch! I've commited this by accident.

keycloak.js
mposolda: Ok, I rewrote that in this manner with the exception of the "post_logout_redirect_url" added just with the "id_token_hint" as mentioned above.

Note that the original version of this "logoutUrl" method did not used the URL/SearchParams and hence I've just continued with using it :-) But agree that this form with URL/SearchParams is slightly cleaner.

mposolda:
@jonkoops Addressed some of your comments and added comment to some points, which I did not addressed in the way you proposed.

keycloak.js
```diff
@@ -400,10 +400,19 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
...
-  '/protocol/openid-connect/logout' +
-  '?redirect_uri=' + encodeURIComponent(redirectUrl);
+Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
+  let url = this.config.realmUrl +
+  '/protocol/openid-connect/logout';
+
+  if (idTokenHint) {
+    let urlParams = new URLSearchParams();
+    urlParams.append("id_token_hint", idTokenHint);
+    if (redirectUrl) {
```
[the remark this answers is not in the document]
jonkoops: Ok, just be aware this is a breaking change in that case. This method is exposed publicly so this would change it's signature.

keycloak.d.ts
```diff
@@ -413,7 +413,7 @@ declare namespace KeycloakConnect {
 
     loginUrl(uuid: string, redirectUrl: string): string
 
-    logoutUrl(redirectUrl: string): string
+    logoutUrl(redirectUrl: string, idTokenHint?: string): string
```
jonkoops, line 416: If the `idTokenHint` is required as mentioned [here](https://github.com/keycloak/keycloak-nodejs-connect/pull/359/files#r837832033) then it should be required at all times. So I assume my previous comment will have to be reverted,
```suggestion
    logoutUrl(redirectUrl: string, idTokenHint: string): string
```

keycloak.js
```diff
@@ -400,10 +400,17 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
...
 };
 
-Keycloak.prototype.logoutUrl = function (redirectUrl) {
-  return this.config.realmUrl +
-  '/protocol/openid-connect/logout' +
-  '?redirect_uri=' + encodeURIComponent(redirectUrl);
+Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
+  const url = new URL(this.config.realmUrl + '/protocol/openid-connect/logout');
+
+  if (idTokenHint) {
```
jonkoops, line 406: We should make both arguments required and remove any defensive coding around it.

jonkoops (changes requested):

stianst:
Does this mean the user has to pass the ID token to initiate the logout? Doesn't the adapter itself have access to that?

keycloak.js
```diff
@@ -400,10 +400,17 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
...
+    url.searchParams.set('id_token_hint', idTokenHint);
+    if (redirectUrl) {
+      url.searchParams.set("post_logout_redirect_uri", redirectUrl);
+    }
+  }
+
+  return url.toString();
 };
```
jonkoops, line 414: Since `redirectUrl` and `idTokenHint` are now mutually exclusive we can reduce nesting a bit by checking them both in a single condition. 

Note that this is a still a breaking change of an existing public API, so this will have to be communicated to consumers of this package.

```suggestion
Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
  const url = new URL(this.config.realmUrl + '/protocol/openid-connect/logout');

  if (redirectUrl && idTokenHint) {
    url.searchParams.set('id_token_hint', idTokenHint);
    url.searchParams.set("post_logout_redirect_uri", redirectUrl);
  }

  return url.toString();
};
```

jonkoops (changes requested):

jonkoops:
> Does this mean the user has to pass the ID token to initiate the logout?

Yes, it does.

keycloak.d.ts
mposolda: Replied in the point below

mposolda:
> > Does this mean the user has to pass the ID token to initiate the logout?
> 
> Yes, it does.

@stianst @jonkoops This is the question as I guess that in most of the cases, people uses Logout handler used for example by the "example" application attached to this repository? This process the logout in the `logout.js`, which itself calls the `logoutUrl` method. And in this case, it is possible to automatically pick the ID token used by the adapter. So for example this "example" did not need to have any updates and it works fine. Also the logout described in the documentation https://www.keycloak.org/docs/latest/securing_apps/index.html#additional-urls does not use `logoutUrl` method directly and hence it would just work fine.

My original assumption was that `logoutUrl` is something like "private" method, which is not supposed to be used by users in their applications at all. I figured that it is kind of "public" when @jonkoops pointed me to the reference in the keycloak.d.ts file. But since it is not mentioned in the documentation or example, I don't know how much this really is a problem...

keycloak.js
mposolda: Yes, this should work.

mposolda: @jonkoops I originally though that by remove "Defensive" you want to remove "if" entirely. Which I think should not be done. But this way is fine.

keycloak.d.ts
mposolda: From my point of view, this is the only open question in this PR whether to change this signature or not. It affects only applications, which directly use `logourUrl` method.

- If we keep ?, it means that developers does not need to update their applications in case they directly use `logoutUrl` method. In this case, users will see "Logout confirmation" screen and will need to confirm logout. So it is change in the behaviour for the end users
- If we don't keep ? and make idTokenHint documented as mandatory parameter, it means that developers will need to update their applications in the case they directly use `logoutUrl` method (which I don't know how much of them is really using as described in the other comment...)

jonkoops:
@mposolda I am ok with this change as it is now as long as it has no adverse effects omitting the `logoutUrl`. Could you do a rebase? I fixed the tests on `main`.

mposolda:
PR to keycloak-documentation: https://github.com/keycloak/keycloak-documentation/pull/1454

pedroigor (approved):

jonkoops:
The CI is failing since this PR depends on changes in Keycloak that are not available in the Docker image that is used for testing. I am working on a fix so that we run the tests against the nightly version instead.

jonkoops:
@mposolda I pulled the changes from here into #364 so we can merge our stuff together.

PR #364, opened 2022-04-05
Test and develop against Nightly versions of Keycloak

jonkoops:
Changes needed for #359. Tests everything against Keycloak Nightly instead of a fixed version Docker image.


mposolda:
@jonkoops In the test results, I can see that tests are failing due the "<p class="instruction">Invalid parameter: redirect_uri</p>" . Which is exactly caused by the OIDC RP-Initiated Logout changes as "redirect_uri" parameter does not work anymore with the logout.

I guess you can cherry-pick the commit from https://github.com/keycloak/keycloak-nodejs-connect/pull/359 to this PR, so both commits are in the same PR. This should work for the tests agains Keycloak 18. WDYT?

mposolda (approved):
@jonkoops Approving as it looks good to me from the first look.

Note that I am not node.js expert, so I am leaving to you whether my review is sufficient to you or whether you rely also on the review of someone else :-)

jonkoops:
Marking this PR as ready for review. One of the tests will fail due to missing changes that will be included in #359.

mposolda (dismissed):

abstractj (approved):
Those changes LGTM, just waiting for the CI.

jonkoops:
I decided to add the changes from #359 here so we can see the full integration. Seems all the tests are passing, so I think we are good to merge.

abstractj (approved):

PR #382, opened 2022-05-24
Fix testing matrix key for Node.js version

jonkoops:


.github/workflows/main.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       # TODO: Split the linting into a separate job.
       - name: Run lint task
         # Only run this step for the default Node version.
-        if: matrix.node == env.DEFAULT_NODE_VERSION
+        if: matrix.node-version == env.DEFAULT_NODE_VERSION
```
mabartos, line 40: You don't need here `${{ matrix.node-version }} == ....` ?

jonkoops: No, everything inside of `if` is considered an expression so there is no need to open an expression here manually.

mabartos: Ok, thanks for the explanation. I was just curious about it.

mabartos (approved):
Approved; even if the CI is broken. It could be done in some follow-up task.

abstractj (dismissed):

PR #419, opened 2022-08-31
Update version on package.json

caiquecastro:


jonkoops:
There is no need for this, we use this version as a placeholder that gets replaced in the release process.

PR #421, opened 2022-09-02
Include redirectUrl even if idTokenHint is not provided

### LINKED ISSUE #420: post_logout_redirect_uri not set by logoutUrl if id hint is missing
### Describe the bug

The `logoutUrl` function on the Keycloak prototype takes to parameters. The first one `redirectUrl` is required, whilst the `idTokenHint` is an optional parameter (per the TS types).

However – if the idTokenHint is not provided, the `redirectUrl` is not included, because of this check 👇 

```js
if (redirectUrl && idTokenHint) {
  url.searchParams.set('id_token_hint', idTokenHint)
  url.searchParams.set('post_logout_redirect_uri', redirectUrl)
}
```

### Version

15.1.1 (server) - 19.0.1 (keycloak-connect)

### Expected behavior

When calling the `logoutUrl` method from the Keycloak prototype, passing a `redirectUrl` I expect the redirectUrl to be included in the logoutUrl returned. Regardless of whether or not the `idTokenHint` is provided.

### Actual behavior

If no `idTokenHint` is passed to the `Keycloak.prototype.logoutUrl` method the `redirectUrl` is not appended either. It's either idTokenHint+redirectUrl or nothing at all.

### How to Reproduce?

```js
const kc = new Keycloak(...)

// this one will include the redirectUrl
kc.logoutUrl('https://where.to.go', 'id123')

// this one won't include the redirectUrl, because of the missing id token hint
kc.logoutUrl('https://where.to.go')
```

### Anything else?

_No response_

kbrabrand:
This PR fixes what I believe is a bug in the keycloak connect client where it will ignore the redirectUrl passed to logoutUrl if no idTokenHint is provided. 

Fixes #420

### DIFF
--- a/keycloak.js
+++ b/keycloak.js
@@ -403,8 +403,11 @@ Keycloak.prototype.loginUrl = function (uuid, redirectUrl) {
 Keycloak.prototype.logoutUrl = function (redirectUrl, idTokenHint) {
   const url = new URL(this.config.realmUrl + '/protocol/openid-connect/logout')
 
-  if (redirectUrl && idTokenHint) {
+  if (idTokenHint) {
     url.searchParams.set('id_token_hint', idTokenHint)
+  }
+
+  if (redirectUrl) {
     url.searchParams.set('post_logout_redirect_uri', redirectUrl)
   }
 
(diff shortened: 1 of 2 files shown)


kbrabrand:
@jonkoops I had trouble getting the tests to run locally and it also seems like github has issues with it. This fixes an issue we have in production. If it looks ok to you and fast-tracking a release of the fix is possible that would be 🌈

PR #449, opened 2022-11-21
Add a deprecation message to the README

jonkoops:
Adds the same deprecation message we've added to the NPM page to the README. This is to avoid folks from getting invested in this library, not knowing of our future plans to remove it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,9 @@
 
 ![Build Status](https://github.com/keycloak/keycloak-nodejs-connect/workflows/CI/badge.svg)
 
+> **Warning**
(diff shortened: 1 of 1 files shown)


edewit (approved):

scoopex:
So, what is the alternative to write a nodejs application which utilizes Keycloak as an auth server?

jonkoops:
We're not ready to make that recommendation yet, but feel free to follow along here https://github.com/keycloak/keycloak-nodejs-connect/issues/453#issuecomment-1349535222.

PR #521, opened 2024-12-10
Create workflows to release Node.js adapter from own repo

### LINKED ISSUE #520: Create workflows to release Node.js adapter from own repo
It can be good to have GH workflow, which will allow to release node.js from it's [own repository](https://github.com/keycloak/keycloak-nodejs-connect) .

NOTE (hint): It looks that there is already script [release.sh](https://github.com/keycloak/keycloak-nodejs-connect/blob/main/release.sh), so perhaps that one is OK to be re-used in the workflow?
I guess this is used by the GH workflow, which is currently triggered from keycloak-rel during keycloak server release (We are going to remove releasing of node.js from keycloak-rel as we want to decouple node.js adapter release from keycloak-server release. Related task for decouple node.js release from keycloak-rel is keycloak/keycloak#35536 )

mposolda:
closes #520

Ability to release node.js adapter from it's own repository. PR introduces 2 workflows:
- `release nightly` - for trigger nightly releases of node.js adapter
- `release` - for trigger regular release of node.js adapter

Workflows are doing same steps like the related workflows in `keycloak-rel`. There is related PR to keycloak-rel, which removes the node.js related workflows from `keycloak-rel` : https://github.com/keycloak-rel-testing/keycloak-rel/pull/45 (So far only PR to the "testing" fork of node.js)

The workflows require 2 secrets present in the repository, which are not there yet:
`GH_TOKEN` - Token of keycloak-bot to be able to use `gh` command (needed for the manipulation with GH releases etc)
`NPM_TOKEN` - Token needed to be able to publish to npm registry like https://www.npmjs.com/package/keycloak-connect/v/26.0.7 . Needed only for `release` (not needed for `release nightly`)

### DIFF
--- a/.github/workflows/x-release.yml
+++ b/.github/workflows/x-release.yml
@@ -1,0 +1,153 @@
+name: X Release
+
+on:
+  workflow_call:
+    inputs:
+      gh-org:
+        required: true
+        type: string
+      branch:
+        required: true
+        type: string
+      tag:
+        required: true
+        type: string
+      nightly:
+        required: true
+        type: boolean
+    secrets:
+      GH_TOKEN:
+        required: true
+      NPM_TOKEN:
+        required: false
+
+concurrency: rel-${{ github.ref }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+
+  show-inputs:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          echo "Github organization: ${{ inputs.gh-org }}  " >> $GITHUB_STEP_SUMMARY
+          echo "Branch: ${{ inputs.branch }}  " >> $GITHUB_STEP_SUMMARY
+          echo "Target tag: ${{ inputs.tag }}  " >> $GITHUB_STEP_SUMMARY
+          echo "Is nightly build: ${{ inputs.nightly }}  " >> $GITHUB_STEP_SUMMARY
+
+  create-tags:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
+
+      - name: Create version commit
+        if: ${{ !inputs.nightly }}
+        run: |
+          ./set-version.sh ${{ inputs.tag }}
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+          git commit -a -m "Set version to ${{ inputs.tag }}"
+
+      - name: Tag commit
+        run: git tag --force ${{ inputs.tag }}
+
+      - name: Push changes
+        run: git push --force origin refs/tags/${{ inputs.tag }}
+
+  create-gh-releases:
+    runs-on: ubuntu-latest
+    needs: [create-tags]
+    env:
+      GH_TOKEN: ${{ secrets.GH_TOKEN }}
+    steps:
+      - name: Check if release exists
+        id: release-exists
+        run: >
+          echo "release-exists=$(
+            if ( gh release view ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect &> /dev/null ); then
+              echo 'true'
+            else
+              echo 'false'
+            fi
+          )" >> $GITHUB_ENV
+
+      - name: Create a release
+        if: env.release-exists == 'false'
+        run: gh release create ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect --title ${{ inputs.tag }} --draft ${{ inputs.nightly && '--prerelease' || '' }}
+
+  release-on-github:
+    runs-on: ubuntu-latest
+    needs: [create-gh-releases]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.tag }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          check-latest: true
+          cache: npm
+
+      - name: Create package tarball
+        run: |
+          npm pack
+          mv -T *.tgz keycloak-nodejs-connect.tgz
+
+      - name: Upload to GitHub Releases
+        run: |
+          for i in `gh release view ${{ inputs.tag }} --json assets --jq '.assets[].name'`; do 
+            test -f $i || gh release delete-asset ${{ inputs.tag }} $i -y
+          done
+          gh release upload ${{ inputs.tag }} keycloak-nodejs-connect.tgz --clobber
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+  publish-gh-release:
+    runs-on: ubuntu-latest
+    needs: [release-on-github]
+    steps:
+      - name: Publish release
+        run: gh release edit ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect --draft=false
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+  publish-npm:
+    runs-on: ubuntu-latest
+    needs: [release-on-github]
+    if: ${{ !inputs.nightly }}
+    steps:
+      - name: Download archive
+        run: gh release download ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect --pattern 'keycloak-nodejs-connect.tgz'
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          check-latest: true
+          registry-url: https://registry.npmjs.org
+
+      - name: Publish to NPM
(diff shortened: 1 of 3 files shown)


.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
+
+      - name: Create version commit
+        if: ${{ !inputs.nightly }}
+        run: |
+          ./set-version.sh ${{ inputs.tag }}
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
jonkoops, line 53: The prefix doesn't really matter here.
```suggestion
          git config --global user.email "github-actions[bot]@users.noreply.github.com"
```

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+          git config --global user.name "github-actions[bot]"
+          git commit -a -m "Set version to ${{ inputs.tag }}"
+
+      - name: Tag commit
+        run: git tag --force ${{ inputs.tag }}
```
jonkoops, line 58: Why are we doing a `--force` here?

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+
+      - name: Push changes
+        run: git push --force origin refs/tags/${{ inputs.tag }}
```
jonkoops, line 61: This would only push the tag right? Don't we also want to push the commit itself?

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: Check if release exists
+        id: release-exists
+        run: >
+          echo "release-exists=$(
+            if ( gh release view ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect &> /dev/null ); then
+              echo 'true'
+            else
+              echo 'false'
+            fi
+          )" >> $GITHUB_ENV
```
jonkoops, line 78: Let's not push this into the `env`, an individual step can have outputs that can be read in other steps (see [GitHub documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter)).

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.tag }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
jonkoops, line 96: Let's use the latest LTS version of Node.js here.
```suggestion
          node-version: 22
```

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+          check-latest: true
+          cache: npm
```
jonkoops, line 98: We're not installing anything, so we don't need the cache.
```suggestion
```

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+    steps:
+      - name: Download archive
+        run: gh release download ${{ inputs.tag }} --repo ${{ inputs.gh-org }}/keycloak-nodejs-connect --pattern 'keycloak-nodejs-connect.tgz'
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
jonkoops, line 136:
```suggestion
          node-version: 22
```

jonkoops (changes requested):
Couple of notes:

1. I think this workflow doesn't need individual jobs, this could all be handled in a single job entirely. This also reduces the amount of re-downloading artifacts and settig up tooling, and makes it all easier to follow.
2. We can probably remove release related scripts such as `set-version.sh` and `release.sh` from the project root.

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+
+concurrency: rel-${{ github.ref }}
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+
+  show-inputs:
```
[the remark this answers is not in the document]
mposolda: +1, copy/pasted it from the `keycloak-rel`, where it is used on many places. But agree not needed and will remove it.

.github/workflows/x-release.yml
mposolda: Again, copy/paste from `keycloak-rel` :-) 

I guess the `--force` is used for the case of broken release, which failed in some further step, but has already tags created. In that case, you are able to re-run the release workflow and eventually overwrite existing tag. But I guess we don't need for the node.js release as it is relatively simple IMO? So will try to remove it.

.github/workflows/x-release.yml
mposolda: Ok, thanks. Was again copy/paste existing stuff from `keycloak-rel` . Will try to improve it.

.github/workflows/x-release.yml
mposolda: The commit itself is pushed just into the tag. But it is not pushed to the `main` branch (or `release/X.Y` branch) itself. I think it is done intentionally that way and it is the approach used in all Keycloak repositories.

For example last commit in the 26.0.7 tag contains the version commit: https://github.com/keycloak/keycloak-nodejs-connect/commits/26.0.7/ .

But the `release/26.0` branch itself, from which the tag was created, does not contain that version commit: https://github.com/keycloak/keycloak-nodejs-connect/commits/release/26.0 .

I prefer to be consistent with other Keycloak repositories and hence keep put the version commit just to the tag. Or did you mean something different?

mposolda:
> Couple of notes:
> 
>     1. I think this workflow doesn't need individual jobs, this could all be handled in a single job entirely. This also reduces the amount of re-downloading artifacts and settig up tooling, and makes it all easier to follow.
> 
>     2. We can probably remove release related scripts such as `set-version.sh` and `release.sh` from the project root.

+1 for the point 1. I will rewrite that to use the steps only and not individual jobs. Was trying to be consistent with `keycloak-rel`, but agree that this could be rather simplified since there is not much "options" needed as in `keycloak-rel` .

For (2), I will remove `release.sh` . For the `set-version.sh`, we should be careful as it is still used by `keycloak-rel` . Hence it can be removed after https://github.com/keycloak-rel/keycloak-rel/pull/137 is merged and once we are sure that none of the `keycloak-rel` builds would be never triggered for the node.js branch from where we remove it. Probably better to remove it in dedicated PR to reduce the dependencies between various PRs? As currently this PR could be merged without the merge of keycloak-rel PR https://github.com/keycloak-rel/keycloak-rel/pull/137 and hence they are not strictly dependent on each other.

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,153 @@
...
+
+  create-tags:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.branch }}
+
+      - name: Create version commit
```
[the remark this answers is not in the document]
mposolda: Updated

.github/workflows/x-release.yml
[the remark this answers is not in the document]
mposolda: Updated

.github/workflows/x-release.yml
mposolda: Updated to remove `--force`

.github/workflows/x-release.yml
mposolda: Updated

mposolda:
@jonkoops Updated a PR to:
- Address most of your inline comments and replied to some remaining
- Using single job with individual steps
- Removed a requirement of `GH_SECRET` as a secret in the repository and instead rely on the automatically created `GITHUB_TOKEN`, which is always available for GH actions (this was requested by @stianst ). Also updated the description of the PR to reflect that
- Removed `release.sh` script, which is not needed

Re-review welcome

.github/workflows/x-release.yml
```diff
@@ -0,0 +1,107 @@
...
+        if: ${{ !inputs.nightly }}
+        run: |          
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+          npm version ${{ inputs.tag }} -m "Set version to ${{ inputs.tag }}"          
+
+      - name: Tag commit
+        run: git tag ${{ inputs.tag }}
```
jonkoops, line 53: By default `npm version` will tag the commit for us, so this step can be skipped for any version that is not the nightly.

jonkoops (changes requested):

.github/workflows/x-release.yml
jonkoops: Hmmm, yeah I guess this makes sense. The question is if we want to do trunk based development for the Node.js adapter and have all the version commits there. I agree that we can keep it like this for now.

jonkoops (approved):
LGTM!

mposolda:
@jonkoops Thanks! I guess we can wait after Keycloak 26.1.0 release (but we can merge beforehand as well if you prefer, but I suppose Keycloak 26.1 release would be still done from `keycloak-rel`). Also the release workflow will need `NPM_SECRET` in this repository, which is not there yet (we will need to ask @stianst for it).

PR #526, opened 2024-12-13
Drop support for end-of-life versions of Node.js

### LINKED ISSUE #525: Drop support for end-of-life versions of Node.js
We still have versions of Node.js in the support target that have been end-of-life for quite some time. This version range wil have to be upgraded to a list of supported Node.js versions, and our support target should be clarified in the README.

jonkoops:
Closes #525

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,13 +5,13 @@ on:
   pull_request:
     branches: [main]
 env:
-  DEFAULT_NODE_VERSION: "18"
+  DEFAULT_NODE_VERSION: "23"
 jobs:
   setup:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ["14", "18"]
+        node-version: ["18", "20", "22", "23"]
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


mposolda:
@jonkoops Thanks! Few questions:

- Do we want to add support for 23 as well? Looks like keycloak-team discussion agreed with 18, 20 and 22, but if you think that 23 is OK as well, we can go with that. If not, we will need to update default version as well (as this PR set it to 23).

- Do we want to add supported versions in the official docs file https://github.com/keycloak/keycloak-nodejs-connect/blob/main/guides/securing-apps/nodejs-adapter.adoc rather than to the README? The official docs file would be rendered in the official Keycloak documentation, so that is likely better place for document this?

mposolda (approved):

jonkoops:
Putting this PR on hold until we figure out why the tests are taking such a long time to run on CI.

rmartinc (approved):
Thanks @jonkoops! LGTM too! And I agree with @mposolda to add the supported versions in a new documentation chapter or paragraph.

jonkoops:
I went ahead and rebased this PR on top of the testing fixes, and now everything passes as expected on newer Node.js versions as well :tada:

PR #611, opened 2025-05-02
Make podman and docker references appear only in upstream

andymunro:
Closes #618

@ahus1 and @pedroigor Can you verify.  I think we don't need a comparable instruction for downstream.


ahus1 (approved):

andymunro:
@ahus1 Is there a way to backport this to 26.2?

ahus1:
I don't see any release branches, and for now we're pulling the documentation from `main` for downstream. Maybe @jonkoops can shed some light on how releases are cut here.
I assume no-one wants to cut a release just for the docs?!

jonkoops:
We currently have not set up any way of doing backports for this project, it was decided that since we'll be getting rid of the Node.js adapter and there has been relatively (essentially no) activity for years, it isn't needed. We can still set it up, but it is probably not worth the effort.

PR #704, opened 2026-05-20
AUT-1352 Skip RSA verification for HS-signed tokens

DmitryMasley:
## Summary
- KC26 signs internal tokens (e.g. `refresh_token`) with HS512 using a realm-side secret that clients never have. `validateToken` currently tries to RSA-verify everything, so every fresh login via KC26 fails with `Grant validation failed. Reason: invalid token (signature)` and the dashboard hits an infinite redirect loop.
- Fix: in [middleware/auth-utils/grant-manager.js](middleware/auth-utils/grant-manager.js#L343-L349), skip RSA verification when `token.header.alg` starts with `HS` and resolve the token as-is. KC validates these tokens server-side when they are presented for refresh.
- Tests added in [test/grant-manager-spec.js](test/grant-manager-spec.js#L453-L517) covering: HS512 + corrupt signature resolves, HS256 + corrupt signature resolves, RS256 + bad signature still rejects (regression guard), and HS-signed token with wrong ISS still rejects (HS shortcut runs after upstream checks).
- Version bump 3.4.41 → 3.4.42.

Ticket: https://smartling.atlassian.net/browse/AUT-1352

## Test plan
- [ ] CI passes `npm run test` against the live Keycloak fixture (existing spec hits `obtainDirectly('test-user', 'tiger')`).
- [ ] Publish `@smartling/keycloak-connect@3.4.42` to Artifactory.
- [ ] Bump `@smartling-express/security` to consume 3.4.42, then verify `tms-dashboard-app` login via KC26 completes without a redirect loop.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,14 +2,11 @@
 *.swo
 *.log
 *.orig
+*.gz
+*.tar
 *.tmp
 node_modules/
-keycloak-*
-!keycloak-*.json
-!keycloak-connect-rest-mixed-client-spec.js
-!keycloak-connect-test.js
-!keycloak-object-test.js
-!keycloak-fixture.json
+/keycloak-*/
 coverage
 pid.txt
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 node_modules/
 keycloak-*
 !keycloak-*.json
+!keycloak-connect-rest-mixed-client-spec.js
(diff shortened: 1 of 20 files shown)


DmitryMasley:
Created in the wrong repo by mistake — re-opening against Smartling/keycloak-nodejs-connect.

